PR #1, opened 2021-04-20
Any: implemented Into conversion for supported types + Display

Horusiath:
This PR add support of convert native Rust data types into Any struct:

- `bool`
- `f32`/`f64`
- `i32`/`i64`
- `u32` (conversion from `u64` not included as it's not safe)
- `String`/`&str`
- `Box<[u8]>`/`Vec<u8>`
- `Option<T>` of convertible type (to support null conversion)
- `Vec<T>` of convertible type (for array)
- `HashMap<String,T>` of convertible type as value

Additionally implemented `std::fmt::Display` trait which is used i.e. in `println!("{}", any_value)`. Printing convert's Any value into JSON-like string.

### DIFF
--- a/lib0/src/any.rs
+++ b/lib0/src/any.rs
@@ -141,10 +141,12 @@ impl TryInto<Any> for u64 {
     type Error = &'static str;
 
     fn try_into(self) -> Result<Any, Self::Error> {
-        if self <= (1 << 53) {
+        if self < (1 << 53) {
             Ok(Any::Number(self as f64))
+        } else if self <= (i64::MAX as u64) {
+            Ok(Any::BigInt(self as i64))
         } else {
-            Err("lib0::Any conversion is possible only for numbers up to 2^53")
+            Err("lib0::Any conversion is possible only for numbers up to 2^63")
         }
     }
 }
--- a/lib0/src/any.rs
+++ b/lib0/src/any.rs
@@ -1,5 +1,6 @@
 use std::cmp::PartialEq;
 use std::collections::HashMap;
+use std::convert::TryInto;
 
 #[derive(PartialEq, Debug, Clone)]
 pub enum Any {
@@ -73,19 +74,13 @@ impl Into<Any> for f32 {
 
 impl Into<Any> for u32 {
     fn into(self) -> Any {
-        Any::BigInt(self as i64)
-    }
-}
-
-impl Into<Any> for i64 {
-    fn into(self) -> Any {
-        Any::BigInt(self)
+        Any::Number(self as f64)
     }
 }
 
 impl Into<Any> for i32 {
     fn into(self) -> Any {
-        Any::BigInt(self as i64)
+        Any::Number(self as f64)
     }
 }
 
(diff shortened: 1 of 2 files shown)


lib0/src/any.rs
```diff
@@ -13,3 +13,131 @@ pub enum Any {
...
+
+impl Into<Any> for f32 {
+    fn into(self) -> Any {
+        Any::Number(self as f64)
+    }
+}
+
+impl Into<Any> for u32 {
+    fn into(self) -> Any {
+        Any::BigInt(self as i64)
```
dmonad, line 76: This should convert to `Any::Number(self as f64)` instead because `Any:Number` is encoded using variable-length encoding and `BigInt` is not. An `u32` fits into `f64`.

lib0/src/any.rs
```diff
@@ -13,3 +13,131 @@ pub enum Any {
...
+    }
+}
+
+impl Into<Any> for i64 {
+    fn into(self) -> Any {
+        Any::BigInt(self)
```
dmonad, line 82: Here, we should check if `self` fits into `Any::Number(f64)`. `f64` can represent `u53`. So you basically just have to check whether `self < 2^53`.

lib0/src/any.rs
```diff
@@ -13,3 +13,131 @@ pub enum Any {
...
+    }
+}
+
+impl Into<Any> for i32 {
+    fn into(self) -> Any {
+        Any::BigInt(self as i64)
```
dmonad, line 88: Similarly this should always fit into `Any::Number`

dmonad (changes requested):
This is a really nice addition to Any encoding and should make it easier for us to do the conversions.

lib0/src/any.rs
Horusiath: If BigInto is not used to store integers (as we cast them to floats whenever possible), what's its purpose?

lib0/src/any.rs
```diff
@@ -73,19 +74,13 @@ impl Into<Any> for f32 {
...
 impl Into<Any> for u32 {
     fn into(self) -> Any {
-        Any::BigInt(self as i64)
-    }
-}
-
-impl Into<Any> for i64 {
-    fn into(self) -> Any {
-        Any::BigInt(self)
+        Any::Number(self as f64)
```
dmonad, line 77: ` i64` doesn't always fit into `f64`. Similar to `u64`, this needs two cases.

lib0/src/any.rs
```diff
@@ -140,4 +135,33 @@ impl<T> Into<Any> for HashMap<String, T> where T: Into<Any> {
         }
         Any::Map(map)
     }
+}
+
+impl TryInto<Any> for u64 {
+    type Error = &'static str;
+
+    fn try_into(self) -> Result<Any, Self::Error> {
+        if self <= (1 << 53) {
```
dmonad, line 144: The largest integer that a double can represent is `2^53-1`. So this needs to be `<`.

lib0/src/any.rs
```diff
@@ -140,4 +135,33 @@ impl<T> Into<Any> for HashMap<String, T> where T: Into<Any> {
...
+            Ok(Any::Number(self as f64))
+        } else {
+            Err("lib0::Any conversion is possible only for numbers up to 2^53")
```
dmonad, line 147: In this case, you can use a `Any::BigInt` instead, as you did before.

dmonad (changes requested):
Any encoding is meant as an efficient alternative to JSON. JSON would represent any number as a string, which is terribly inefficient.

`Any` will write integers using variable-length encoding (possibly less than 1 byte per integer if the number is less or equal than 127). Other integers are also encoded very efficiently. 

While `JSON` encoding can only represent doubles. `Any` can also represent BigInt, which is now also available in JavaScript. 

So all of this might seem weird for someone who is used to hyper efficiency of a native language. In a scripting language, a number is either a double or a BigInt. So this type of encoding is (I hope so) a big performance improvement that can happen in the background without the user noticing.

lib0/src/any.rs
Horusiath: Github display broke here. It's `impl Into<Any> for u32`.

lib0/src/any.rs
dmonad: Oh, didn't see this comment. It's just that `Any::Number` should be preferred. For i64, sure use `Any::BigInt` if you use more than 53 bits.

dmonad (approved):

PR #3, opened 2021-04-21
Put some duck tape to make the project compile again

Horusiath:
In general I've seen that a lot of encoding/decoding logic here is inlined into calling methods. I think we could optimize it into a kind of `Encode`/`Decode` traits and just implement them for necessary cases eg.:

```rust
trait Encode {
  fn encode(&self, encoder: &mut Encoder);
}

trait Decode {
  fn decode(decoder: &mut Decoder) -> Result<Self>;
}

impl Encode for Item { ... }
impl Decode for Item { ... }
```

This is very popular pattern, most commonly known in [serde](https://serde.rs/), which is the facto standard for serialization/deserialization in Rust ecosystem. Re. performance, it's also pretty fast: Rust resolves correct function calls at compile time.

### DIFF
--- a/yrs/src/block_store.rs
+++ b/yrs/src/block_store.rs
@@ -100,7 +100,7 @@ impl BlockStore {
         for i in 0..num_of_state_updates {
             let number_of_structs = update_decoder.rest_decoder.read_var_uint::<u32>() as usize;
             let client = update_decoder.read_client();
-            let clock: u32 = update_decoder.rest_decoder.read_var_uint();
+            let mut clock: u32 = update_decoder.rest_decoder.read_var_uint();
             let structs = store.get_client_structs_list_with_capacity(client, number_of_structs as usize);
             let id = block::ID { client, clock };
             for j in 0..number_of_structs {
--- a/yrs/src/block_store.rs
+++ b/yrs/src/block_store.rs
@@ -4,6 +4,7 @@ use lib0::decoding::Decoder;
 use lib0::encoding::Encoder;
 use std::collections::HashMap;
 use std::vec::Vec;
+use crate::block::Block;
 
 impl StateVector {
     pub fn empty() -> Self {
--- a/yrs/src/block_store.rs
+++ b/yrs/src/block_store.rs
@@ -90,13 +90,14 @@ impl BlockStore {
     pub fn new() -> Self {
         Self {
             clients: HashMap::<u64, ClientBlockList, BuildHasherDefault<ClientHasher>>::default(),
+            local_block_list: ClientBlockList::new(),
         }
     }
-    pub fn from (update_decoder: updates::decoder::DecoderV1) -> Self {
+    pub fn from (update_decoder: &mut updates::decoder::DecoderV1) -> Self {
         let mut store = Self::new();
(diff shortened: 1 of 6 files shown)


yrs/src/block_store.rs
```diff
@@ -72,8 +73,11 @@ impl ClientBlockList {
         if self.integrated_len == 0 {
             0
         } else {
-            let item = &self.list[self.integrated_len - 1];
-            item.id.clock + 1
+            match &self.list[self.integrated_len - 1] {
+                Block::Item(item) => item.id.clock + 1,
+                Block::Skip(skip) => skip.len + 1,
+                Block::GC(gc) => gc.len + 1,
```
Horusiath, line 79: I used intuition here, based on how `Store::from` works. I have no other reason to think this is correct.

dmonad: The state of a client is basically the next expected id from a certain client. 

So assume you apply item `{ id: { client: 0, clock: 0 }, len: 1 }` and then  `{ id: { client: 0, clock: 1 }, len: 2 }`. The next expected item from client 0 is ` { id: { client: 0, clock: 3 }, len: _ }`. 

So after applying the first operation, the state is `1`. After applying the next operation (which has clock = 1), the state is 3. After applying the third operation, the state is 3 + len (depending on the length of the operation). 

The length of an operation is determined by the amount of content an item fits. I use a compound representation to represent several insertions in a single item.

We should talk about the idea of compound representation of operations. I talked about my motivation behind them in my blog: https://blog.kevinjahns.de/are-crdts-suitable-for-shared-editing/

This should be `item.id.clock + item.len()`, and `skip.id.clock + skip.len()`, and `gc.id.clock + gc.len()` respectively.

yrs/src/lib.rs
```diff
@@ -115,9 +115,9 @@ pub struct ClientBlockList {
     pub integrated_len: usize,
 }
 
-
 pub struct BlockStore {
     pub clients: HashMap<u64, ClientBlockList, BuildHasherDefault<ClientHasher>>,
+    pub local_block_list: ClientBlockList,
```
dmonad, line 120: I was just about to get rid of `local_block_list`. The idea was that the local client would not have to access `clients.get(local_client_id)` to get to the blocklist. But it turned out that this produces too many cases.

dmonad: No need to get rid of it now as long as it compiles. But maybe you could add a `todo` notice that we want to remove this.

yrs/src/block_store.rs
```diff
@@ -130,10 +135,19 @@ impl BlockStore {
                     } else {
                         None
                     };
-                    block::ItemContent::Binary()
+                    let item: block::Item = todo!();
+                    structs.list.push(block::Block::Item(item));
+                    clock += 1;
```
dmonad, line 140: This should be `clock += item.len()`

dmonad (changes requested):
Thanks for getting this to work again! 

With the information you have you did a terrific job understanding the codebase. In a couple of places I requested changes that I'd like to change before merging, so we don't have to debug the code that is not working.

dmonad (approved):

PR #7, opened 2021-04-21
adding uncommitted work

dmonad:
I had some uncommitted work on my computer (WIP). We don't have to merge this because some of it was already addressed by #3. This implements, for example, the `len()` method on generic blocks.  @Horusiath

### DIFF
--- a/yrs/src/block_store.rs
+++ b/yrs/src/block_store.rs
@@ -53,21 +53,18 @@ impl StateVector {
 }
 
 impl ClientBlockList {
-    #[inline]
     fn new() -> ClientBlockList {
         ClientBlockList {
             list: Vec::new(),
             integrated_len: 0,
         }
     }
-    #[inline]
     pub fn with_capacity(capacity: usize) -> ClientBlockList {
         ClientBlockList {
             list: Vec::with_capacity(capacity),
             integrated_len: 0,
         }
     }
-    #[inline]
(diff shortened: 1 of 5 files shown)


Horusiath (approved):
Looks ok for me. If you'll merge it, I'll rebase my branch on top of it.

yrs/src/block_store.rs
```diff
@@ -173,11 +173,16 @@ impl BlockStore {
         }
     }
     #[inline(always)]
+    pub fn get_block(&self, ptr: &block::BlockPtr) -> &block::Block {
+        &self.clients[&ptr.id.client].list[ptr.pivot as usize]
+    }
+    #[inline(always)]
```
Horusiath, line 179: In general, I would put inline directives at the very end of the development process. LLVM can do real miracles on it's own and forcing some habits on it can sometimes screw things up instead of making the faster eg. if inlined function causes function calling it to be too big to be inlined of contains some logic that prevents inlining (we already noticed this behavior in .NET where throw exception prevented callee method from being inlined... it was literally faster to wrap exceptions into methods marked to never be inlined).

dmonad: I added these comments when I did initial benchmarks using the BlockStore approach. It improved performance by 10% just because the test did only lookups. I think you are right that we should remove these comments and add them later based on real benchmarks.

PR #8, opened 2021-04-22
Implement find_pivot and IdSet

Horusiath:
This PR is build on top of #3. I need a stable compiling branch and it was the one.

Work in progress: This PR introduces implemented `find_pivot` (equivalent of `findIndexSS` from JS API as I presume) and most of the `IdSet` (again as I assume it was supposed to be equivalent of `DeleteSet`).

- IdSet iteration is still missing - it requires adding some extra stuff to `Transaction` which I'm still figuring out.
- I want to introduce tests for implemented functions, but still need to figure that part out. Maybe I'll have to postpone them until a Doc API will be more complete. We'll see.

### DIFF
--- a/yrs/src/block.rs
+++ b/yrs/src/block.rs
@@ -23,6 +23,15 @@ pub struct ID {
     pub clock: u32,
 }
 
+impl ID {
+    pub fn new(client: u64, clock: u32) -> Self {
+        ID {
+            client,
+            clock,
+        }
+    }
+}
+
 #[derive(Copy, Clone)]
 pub struct BlockPtr {
     pub id: ID,
@@ -104,6 +113,33 @@ impl Block {
             }
         }
     }
+
+    pub fn id (&self) -> &ID {
+        match self {
+            Block::Item(item) => {
+                &item.id
+            }
+            Block::Skip(skip) => {
+                &skip.id
+            }
+            Block::GC(gc) => {
+                &gc.id
+            }
+        }
+    }
+    pub fn len (&self) -> u32 {
+        match self {
+            Block::Item(item) => {
+                item.content.len()
+            }
+            Block::Skip(skip) => {
+                skip.len
+            }
+            Block::GC(gc) => {
+                gc.len
+            }
+        }
+    }
 }
 
 pub enum ItemContent {
@@ -116,6 +152,43 @@ pub enum ItemContent {
     Format(String, String), // key, value: JSON
     String(String),
     Type(types::Inner),
+}
+
+impl ItemContent {
+    pub(crate) fn splice(&mut self, offset: usize) -> Option<ItemContent> {
+        match self {
+            ItemContent::Any(value) => {
+                todo!()
+            },
+            ItemContent::String(string) => {
+                let (left, right) = string.split_at(offset);
+                let mut left = left.to_string();
+                let mut right = right.to_string();
+
+                //TODO: do we need that in Rust?
+                //let split_point = left.chars().last().unwrap();
+                //if split_point >= 0xD800 as char && split_point <= 0xDBFF as char {
+                //    // Last character of the left split is the start of a surrogate utf16/ucs2 pair.
+                //    // We don't support splitting of surrogate pairs because this may lead to invalid documents.
+                //    // Replace the invalid character with a unicode replacement character (� / U+FFFD)
+                //    left.replace_range((offset-1)..offset, "�");
+                //    right.replace_range(0..1, "�");
+                //}
+                *self = ItemContent::String(left);
+
+                Some(ItemContent::String(right))
+            },
+            ItemContent::Deleted(len) => {
+                let right = ItemContent::Deleted(*len - offset as u32);
+                *len = offset as u32;
+                Some(right)
+            },
+            ItemContent::JSON(value) => {
+                todo!()
+            },
+            _ => None,
+        }
+    }
 }
 
 #[derive(Clone)]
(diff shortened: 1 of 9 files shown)


dmonad:
I merged #3, but lets try to remember to change [this](https://github.com/yjs/y-crdt/pull/3/files#diff-f7b7b91df1195b2117df73e35947f3d10b3169909f95eb963f12a132ea8e0aa4R134) to `clock += item.len()`.

yrs/src/block.rs
```diff
@@ -39,6 +50,107 @@ pub enum Block {
...
+            Block::Item(item) => Some(item),
+            _ => None
+        }
+    }
+
+    pub fn is_deleted(&self) -> bool {
+        match self {
+            Block::Item(item) => item.deleted,
+            Block::Skip(_) => false,
+            Block::GC(_) => false,
```
dmonad, line 72: `GC` is a garbage-collected deleted operation. So this should return true.

yrs/src/block.rs
```diff
@@ -51,6 +163,43 @@ pub enum ItemContent {
...
+        match self {
+            ItemContent::Any(value) => {
+                todo!()
+            },
+            ItemContent::String(string) => {
+                let (left, right) = string.split_at(offset);
+                let mut left = left.to_string();
+                let mut right = right.to_string();
+
+                //TODO: do we need that in Rust?
```
dmonad, line 177: The approach in Yjs is that it is not allowed to split surrogate pairs. If you want to change an emoji from :smile:  to :frowning:, then you must replace the whole emoji even if they start with the same character. Fortunately, this is what editors in Yjs already do. If we ever run into issues, we should prevent splitting surrogate pairs with an error message.

yrs/src/block.rs
```diff
@@ -101,9 +250,29 @@ impl Item {
...
     pub fn len (&self) -> u32 {
         self.content.len()
     }
+
+    pub fn split(&mut self, diff: u32) -> Item {
+        let client = self.id.client;
+        let clock = self.id.clock;
+        let other = Item {
+            id: ID::new(client, clock + diff),
+            left: Some(BlockPtr::from(self.id)),
```
dmonad, line 263: This should point to the end of `self` after the split. So left is `left: Some(BlockPtr::from(ID::new(client, clock + diff - 1)))`

yrs/src/id_set.rs
```diff
@@ -1,23 +1,200 @@
...
+            let mut ranges = Vec::with_capacity(blocks.list.len());
+            let mut i = 0;
+            while i < blocks.list.len() {
+                let block = &blocks.list[i];
+                if block.is_deleted() {
+                    let clock = block.id().clock;
+                    let mut len = block.len();
+                    if i + 1 < blocks.list.len() {
+                        let mut next = &blocks.list[i+1];
+                        while i + 1 < blocks.list.len() && next.id().clock == clock + len && next.is_deleted() {
```
dmonad, line 67: I think we can remove this condition `next.id().clock == clock + len`. This should always hold true.

yrs/src/block.rs
```diff
@@ -15,13 +17,22 @@ const BLOCK_ITEM_ANY_REF_NUMBER: u8 = 8;
 const BLOCK_ITEM_DOC_REF_NUMBER: u8 = 9;
 const BLOCK_SKIP_REF_NUMBER: u8 = 10;
 
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, PartialOrd, PartialEq, Ord, Eq, Hash)]
```
dmonad, line 20: I want to be a bit careful about whether a Hash function is applied on `TypePtr` / `BlockPtr` / `ID`. 

Is the `Hash` needed to do comparisons? Can we optimize this using a comparison `id.client == other.client && id.clock === other.clock`. Possibly we can improve that by comparing the 12 bytes directly.

We could compute ordering as `id < other ⇔ id.client < other.client  ||  (id.client == other.client && id.clock < other.clock)` (which is a common comparison method used in Yjs). Then we could reuse ordering in the integration process.

dmonad (changes requested):

yrs/src/block.rs
Horusiath: 👍  I did that but apparently erased while cleaning up commit.

yrs/src/id_set.rs
Horusiath: Had it ever failed in y.js? I've ported it from there. Usually there's a reason for such condition to happen in first place.

yrs/src/block.rs
Horusiath: Had to do that because transaction.changes is a map which uses `TypePtr` as a key. I can probably use dedicated Hasher instead.

Horusiath: Re operators:
- I removed `PartialOrd` and `Ord` as we usually don't need them.
- `PartialEq` and `Eq` are already using structural comparison and optimizations like memcmp will be usefull, I'd leave them for later.
- For `Hash` I've implemented `XorHasher` (which simply builds hash as series of xor transforms), but again: I'd leave these sorts of optimizations till the end, when we'll have some benchmarks and not apply extra logic in blind.

Horusiath:
Feel free to merge this if there are no objections left. I'll continue working on test cases on separate branch.

yrs/src/id_set.rs
dmonad: Previous versions of Yjs didn't always require that all operations are present. So you could have `id={clock: 1}` but `id={clock: 0}` could still be missing. I think this is reminiscent. 

I just removed it from Yjs and the test suite is still running fine. 
https://github.com/yjs/yjs/commit/4e343ccace9dca0b594ab42d5934b43e95d245c6

The CI that runs the complete test suite is still running. I'll ping here if it breaks.

PR #11, opened 2021-05-01
Project formatting + work over the block store decoding

Horusiath:
This is big piece mostly because I put some rust fmt to provide a consistent formatting across the project. Other changes include:

1. Removing unused imports .
2. Switching from camelCase to snake_case convention for names that didn't followed it.
3. Moving struct definitions and implementations together into the same file (eg. `Doc` to doc.rs, `Transaction` to transaction.rs, `Store` to store.rs, `BlockStore` to block_store.rs etc.) - it's a standard practice in Rust. Additionally I've exposed these structures at the top level in `lib.rs`.
4. Replaced hardcoded numbers with constants and renamed constants to use more meaningful names eg. `BIT8` &rarr; `HAS_ORIGIN`.
5. Replaced fancy `TypeRefs` struct with alias `type TypeRefs = u8`.
4. Last but not least, I've finished implementation of `BlockStore::from` function and provided a basic test, which proves that it can decode the state encoded by yjs. It's the most atomic test I could think of. We could probably make use of another one which is much more elaborate and tests most of the possible branches, but I still need to think about possible ways to assert correct behavior.


yrs/src/store.rs
```diff
@@ -1,58 +1,205 @@
...
+        let local_block_list = self.blocks.get_client_blocks_mut(self.client_id);
+        local_block_list.list.push(block::Block::Item(item));
+        local_block_list.integrated_len += 1;
+    }
+
+    pub fn read_blocks(&mut self, update_decoder: &mut updates::decoder::DecoderV1) {
+        let number_of_clients: u32 = update_decoder.rest_decoder.read_var_uint();
+        for _ in 0..number_of_clients {
+            let client = update_decoder.read_client();
+            let number_of_structs: u32 = update_decoder.rest_decoder.read_var_uint();
```
Horusiath, line 110: Adding latest test (`apply_update_basic`) confirmed that this step doesn't work as expected. Number of structs assigned by yjs should be 3, but upon reading its 1490905955.

PR #13, opened 2021-05-03
Event handlers

Horusiath:
Depends on #11

Adds a generic `EventHandler<T>` type which enables multiple subscribers.


yrs/src/block.rs
```diff
@@ -275,109 +220,126 @@ impl Item {
...
+                let right = right.to_vec();
+                *self = ItemContent::Any(left);
+                Some(ItemContent::Any(right))
             }
+            ItemContent::String(string) => {
+                let (left, right) = string.split_at(offset);
+                let left = left.to_string();
+                let right = right.to_string();
+
+                //TODO: do we need that in Rust?
```
dmonad, line 321: There were several discussions around this in Yjs. The best solution seems to be to just disallowing splitting surrogate pairs. The Yjs algorithm is not prone to interleaving, so this should be fine.

FYI: However, what we should do - at some point - is to switch to utf16 strings because that is what is used in Yjs. At least the length of a string `item.len()` should be computed based on the utf16 len, otherwise the algorithm will break when complex characters are inserted.

yrs/src/block_store.rs
```diff
@@ -125,54 +135,68 @@ impl BlockStore {
...
+                        } else {
+                            types::TypePtr::Id(block::BlockPtr::from(update_decoder.read_left_id()))
+                        };
+                        let parent_sub = if cant_copy_parent_info && (info & 0b00100000 != 0) {
+                            Some(update_decoder.read_string().to_owned())
+                        } else {
+                            None
+                        };
+                        let content =
+                            ItemContent::decode(update_decoder, info, BlockPtr::from(id.clone())); //TODO: What BlockPtr here is supposed to mean
```
dmonad, line 185: This is a reference to the `item` itself. If this is a type, we need some kind of reference to the type. A `TypePtr` will point to an `item` that contains a type.

yrs/src/event.rs
```diff
@@ -0,0 +1,89 @@
+use rand::RngCore;
```
dmonad, line 1: Sweet. If you plan to integrate this, could we maybe do that as an outside layer outside of `yrs`? The language bindings certainly don't need that and should just observe changes based on the dropped transactions. On the other hand, we need an API for Rust code.

Maybe it would make sense to separate the efforts a little bit more. We could have..

* `y-base-crdt` as the basis implementation of the CRDT (what `yrs` is right now)
* `y-binding-wrapper` as the wrapper around `y-base-crdt` that can be easily exported to other languages. This doesn't expose a lot of internal information.
* `yrs` The Rust port of Yjs, which simply includes `y-base-crdt` and provides a nice idiomatic Rust API.
* `ypy` Python port
* `ygo` Go port
* ...

WDYT?

dmonad (approved):
So much is happening!

It looks good. I'm going to merge this. 

I added a couple of explanations & questions below.

yrs/src/event.rs
Horusiath: I'm still thinking about the best way of project organization. So far I think that we could have `y-rs` crate as a core implementation that might be also Rust-idiomatic. I'm curious to try if C FFI bindings (or even WASM bindings) could be just realized as [feature flags](https://doc.rust-lang.org/cargo/reference/features.html) to that crate.

But for now I wouldn't put that into highlight. Originally we planed to make crates in 3rd milestone. Reorganizing code around is cheap when done separately and not mixed with adding other features (like in this branch xD).

We should keep in mind that bindings cannot have generic and lifetime parameters in extern languages APIs and to make sure that dependencies we use can work with WASM/C (eg. `rand` crate worked with WASM but only to version 0.6.3, later ones didn't).

PR #16, opened 2021-05-12
[WIP] Integration function

Horusiath:
Work in progress. Build on top PR #15 (which could be merged first).

Some ideas after talk with @dmonad - we could split integration function into steps:

1. Don't try to decode update payload into `BlockStore` directy - use separate data structure optimized for this task.
2. Instead of resolving dependency chain in main loop of integration process, set them already in the right order (predecessor->successor) during decoding.
3.  When check if any of the existing items need to be split in order to integrate the update and do it next.
4. Once all update operations are in right order and all blockstore items are split, only then apply operations on already prepared block store.

Another thing we talked about is related to finding the gaps in updates and involve `IdSet` (delete set) for that instead of state vectors. `IdSet` is aligned with the concept described as [Dotted Version Vector](https://riak.com/posts/technical/vector-clocks-revisited-part-2-dotted-version-vectors/index.html?p=9929.html) and should fit this idea well.

### DIFF
--- a/lib0/benches/lib0_benchmarks.rs
+++ b/lib0/benches/lib0_benchmarks.rs
@@ -12,11 +12,11 @@ fn criterion_benchmark(c: &mut Criterion) {
         b.iter(|| {
             let mut encoder = Encoder::new();
             for i in 0..(BENCHMARK_SIZE as i64) {
-                encoder.write_var_int(i);
+                encoder.write_ivar(i);
             }
             let mut decoder = Decoder::new(&encoder.buf);
             for i in 0..(BENCHMARK_SIZE as i64) {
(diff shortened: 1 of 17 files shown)


Horusiath:
I was thinking about inefficiency of splitting integration procedure into separate steps - creating up the dependency chain, splitting up items then integrating updates. In theory this means that we might need to traverse the update blocks more times than strictly necessary, which is significant for eg. loading document which potentially can consist of millions of blocks. However I think there's an optimization step for this case - **it's safe to assume that blocks within the update itself have already been split in a correct way**, so we only may need to split blocks already existing in the block store prior the update.

PR #17, opened 2021-05-20
Integrate and merge functions

Horusiath:


Horusiath:
@dmonad I fixed the issues revealed by the tests so far, there are few of them so far, but they are all passing now.

yrs/src/block.rs
```diff
@@ -51,7 +60,7 @@ impl BlockPtr {
     }
 }
 
-#[derive(Debug, PartialEq)]
+#[derive(Debug, Clone, PartialEq)]
```
Horusiath, line 63: Atm. integration process does deep cloning. It's because Rust's `Vec`/`VecDeque` only allow to take elements from back/front. I've got an idea that's basically something like non-resizable vec, that tracks the occupation of the indexes, this way allowing us to remove elements from the middle of it. I'll add it in the separate PR.

PR #21, opened 2021-06-11
Text insert/delete with tests

Horusiath:
This PR implements features of update API related to deleting blocks - with regards to possible block slicing and applying updates.


yrs/src/block.rs
```diff
@@ -781,6 +806,73 @@ impl ItemContent {
     }
 }
 
+impl std::fmt::Display for ID {
```
Horusiath, line 809: I've implemented Display traits for core components, so that we can print out BlockStore contents in more compacts and comprehensible way.

PR #22, opened 2021-06-21
Ability to subscribe to incoming updates

Horusiath:
Merge after #21

This PR used event handlers defined in the past and incorporates them into publish/subscribe for `update` application event. For sample scenario, see test.

### DIFF
--- a/yrs/src/doc.rs
+++ b/yrs/src/doc.rs
@@ -90,7 +90,7 @@ mod test {
     use crate::updates::decoder::Decode;
     use crate::updates::encoder::{Encode, Encoder, EncoderV1};
     use crate::{Doc, StateVector};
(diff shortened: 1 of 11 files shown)


yrs/src/doc.rs
```diff
@@ -157,4 +171,32 @@ mod test {
...
+        let counter = Rc::new(Cell::new(0));
+        let doc = Doc::new();
+        let mut doc2 = Doc::new();
+        let c = counter.clone();
+        let sub = doc2.on_update(move |e| {
+            for block in e.update.blocks() {
+                c.set(c.get() + block.len());
+            }
+        });
+        let mut txn = doc.transact();
```
Horusiath, line 186: This test presents how to subscribe to incoming updates.

PR #25, opened 2021-06-24
More tests for binary compatibility between Yjs and Yrs

Horusiath:
This PR adds additional tests to check if we can properly deserialize and serialize back again following updates regarding following types:

- YArray
- YMap
- YXmlFragment with YXmlElement

These tests are not checking if integration process was successful (it's a separate step), but only if binary payload can be read/written in format compatible with Yjs.


dmonad (approved):

Horusiath:
@dmonad I think you need to approve it again.

PR #30, opened 2021-07-19
YMap implementation

Horusiath:
Missing: tests - I think it would be a good idea to implement them directly on top of simulation framework (PR #26).


Horusiath:
@dmonad YMap works in principle - I think this PR is ready for review.

PR #37, opened 2021-08-30
[WIP] Port merge_updates from Yjs

dmonad:


dmonad:
@Horusiath Finished with implementing your suggestions and I also rebased to the latest main.

PR #39, opened 2021-09-07
Documentation and encapsulation of the API

Horusiath:
Merge after #38

This PR contains rust docs for public API methods. It also encapsulates methods and types that do not need to be exposed to end users.

It also applies some of the naming convention changes:

- `Inner`/`InnerRef` renamed to `Branch`/`BranchRef`. Reason: inner is too generic and cannot be used to any sensible explanation (also for the context of future articles describing architecture). Since document can be represented as a tree of nodes, that are either leafs (items) or branches (inners) I thought that a tree "branch" is the best way to describe the content.
- All methods used generating binary contents (like generation of updates and state vectors binaries) are suffixed with `_v1`. Since rust doesn't allow method name overloading, and we know that v2 exists, it's better to make a versioning of methods using decoders explicit.
- Concepts of compact/merge (used to join and pack metadata from two different structures into one) are renamed to `squash`. The name was taken from git's squash function which is well recognized and conceptually similar (instead of adjacent commits, here we're squashing adjacent blocks or delete set ranges).


yrs/src/block.rs
```diff
@@ -1,36 +1,70 @@
...
 pub const HAS_RIGHT_ORIGIN: u8 = 0b01000000;
+
+/// Bit flag used to tell if encoded item has left origin defined.
 pub const HAS_ORIGIN: u8 = 0b10000000;
+
+/// Bit flag used to tell if encoded item has a parent subtitle defined. Subtitles are used only
+/// for blocks which act as map-like types entries.
 pub const HAS_PARENT_SUB: u8 = 0b00100000;
 
+/// Unique block identifier.
```
dmonad, line 58: Also say that this is a Lamport Timestamp. 

> The latest ID created by each user forms the state vector.

yrs/src/block.rs
```diff
@@ -765,6 +894,8 @@ impl ItemContent {
         }
     }
 
+    /// Checks if item content can be considered countable. Countable elements contribute to
+    /// a length of the block they are contained by.
```
dmonad, line 898: The block length is not affected by is_countable. Describe the difference between block-length and length of the shared type

yrs/src/block_store.rs
```diff
@@ -10,44 +10,62 @@ use std::hash::BuildHasherDefault;
 use std::ops::{Index, IndexMut};
 use std::vec::Vec;
 
+/// State vector is a compact representation of all known blocks inserted and integrated into
+/// a given document. This descriptor can be serialized and used to determine a difference between
+/// seen and unseen inserts of two replicas of the same document, potentially existing in different
+/// processes.
```
dmonad, line 16: Link to version vector?

dmonad (changes requested):

dmonad (approved):

PR #43, opened 2021-11-05
[WIP] Fuzzy tests

Horusiath:


yrs/src/block.rs
```diff
@@ -419,7 +459,7 @@ pub(crate) struct Item {
     pub parent_sub: Option<String>, //TODO: Rc since it's already used in Branch.map component
 
     /// Bit flag field which contains information about specifics of this item.
-    pub info: Cell<u8>,
+    pub info: u8,
```
Horusiath, line 462: Since block store now accesses `Block`s via `UnsafeCell`, we no longer need cells here. We can just obtain mutable block reference at any point in time.

yrs/src/block.rs
```diff
@@ -527,10 +567,9 @@ impl Item {
         if let Some(origin_id) = self.origin {
             let ptr = BlockPtr::from(origin_id);
             if let Some(item) = txn.store.blocks.get_item(&ptr) {
-                let id = self.origin.unwrap();
                 let len = item.len();
-                if id.clock == item.id.clock + len - 1 {
-                    self.left = Some(ptr);
+                if origin_id.clock == item.id.clock + len - 1 {
```
Horusiath, line 571: `origin_id` and `id` point to the same block but are not necessarily equal - if block length is > 1, then these two can easily differ.

yrs/src/block.rs
```diff
@@ -594,7 +633,7 @@ impl Item {
             if let Some(mut left) = left {
                 if let Some(origin) = txn.store.blocks.get_item(&left) {
                     self.origin = Some(origin.last_id());
-                    left.id = origin.last_id();
+                    left.id = origin.id;
```
Horusiath, line 636: left/right pointers should always point to the beginning of the neighbor block (for the sake of block store equality). left/right origin IDs point to actual position in a timeline set by Lamport Timestamps - for this reason left origin points to `last_id` while left neighbor points to id`.

yrs/src/block.rs
```diff
@@ -750,8 +789,16 @@ impl Item {
...
-            let parent_deleted = self.is_deleted();
+            txn.add_changed_type(&*parent_ref, self.parent_sub.as_ref());
+            let parent_deleted = if let TypePtr::Id(ptr) = &self.parent {
+                if let Some(item) = txn.store.blocks.get_item(ptr) {
+                    item.is_deleted()
+                } else {
+                    true
+                }
+            } else {
+                false
```
Horusiath, line 800: This is accurate representation of corresponding Yjs code - previous implementation used simplification.

yrs/src/block.rs
```diff
@@ -808,7 +855,7 @@ impl Item {
         let clock = self.id.clock;
         let other = Item {
             id: ID::new(client, clock + diff),
-            left: Some(BlockPtr::from(ID::new(client, clock + diff - 1))),
+            left: Some(BlockPtr::from(self.id)),
```
Horusiath, line 858: Again - while left origin points to the end of a previous block, left pointer point to the beginning of it.

yrs/src/block_store.rs
```diff
@@ -398,25 +451,39 @@ impl BlockStore {
 
     /// Returns mutable reference to an item, given its pointer. Returns `None` if not such block
     /// could be found.
-    pub(crate) fn get_item_mut(&mut self, ptr: &block::BlockPtr) -> Option<&mut block::Item> {
-        let blocks = self.clients.get_mut(&ptr.id.client)?;
-        let block = blocks.list.get_mut(ptr.pivot())?;
+    pub(crate) fn get_item_mut(&self, ptr: &block::BlockPtr) -> Option<&mut block::Item> {
+        let block = self.get_block_mut(ptr)?;
```
Horusiath, line 455: Previous implementation was constrained to use only `BlockPtr::pivot` field - pivots are not to be trusted and sometimes we may need to fallback using standard search.

yrs/src/id_set.rs
```diff
@@ -454,8 +456,33 @@ impl DeleteSet {
         self.0.squash()
     }
 
-    pub(crate) fn try_compact(&mut self, blocks: &BlockStore) {
-        //TODO
+    pub(crate) fn try_squash_with(&mut self, store: &mut Store) {
+        // try to merge deleted / gc'd items
+        for (client, range) in self.iter() {
```
Horusiath, line 461: Previously process of squashing block store using delete set was not implemented, resulting in differing internal structure of documents after applying local/remote updates.

yrs/src/test_utils.rs
```diff
@@ -31,12 +30,18 @@ const MSG_SYNC_STEP_1: usize = 0;
 const MSG_SYNC_STEP_2: usize = 1;
 const MSG_SYNC_UPDATE: usize = 2;
 
-pub fn run_scenario<F>(mods: &[F], users: usize, iterations: usize)
+pub fn run_scenario<F>(mut seed: u64, mods: &[F], users: usize, iterations: usize)
 where
-    F: Fn(&mut Doc, &mut ThreadRng),
+    F: Fn(&mut Doc, &mut StdRng),
```
Horusiath, line 35: Using `StdRng` since `ThreadRng` API doesn't expose seed initialization.

yrs/src/types/map.rs
```diff
@@ -120,11 +120,7 @@ impl Map {
     pub fn clear(&self, txn: &mut Transaction<'_>) {
         let t = self.0.borrow();
         for (_, ptr) in t.map.iter() {
-            if let Some(item) = txn.store.blocks.get_item(ptr) {
-                if !item.is_deleted() {
-                    item.mark_as_deleted();
-                }
-            }
+            txn.delete(ptr);
```
Horusiath, line 123: Use full deletion process on clear, which does more than just setting tombstone flag.

yrs/src/update.rs
```diff
@@ -169,7 +189,7 @@ impl Update {
             while let Some(mut block) = stack_head {
                 let id = block.id();
                 if local_sv.contains(id) {
-                    let offset = local_sv.get(&id.client) - id.clock;
+                    let offset = local_sv.get(&id.client) as i32 - id.clock as i32;
```
Horusiath, line 192: This can go negative.

yrs/src/block_store.rs
```diff
@@ -172,43 +218,46 @@ impl ClientBlockList {
...
     /// Returns a mutable block reference, given a pointer identifier of that block.
     /// Returns `None` if no block with such reference could be found.
     pub(crate) fn find(&mut self, ptr: &BlockPtr) -> Option<&mut Block> {
-        let pivot = match self.list.get_mut(ptr.pivot()) {
-            Some(block) if *block.id() == ptr.id => Some(ptr.pivot()),
-            _ => self.find_pivot(ptr.id.clock),
-        };
-        self.list.get_mut(pivot?)
+        match self.try_get_mut(ptr.pivot()) {
+            Some(block) if *block.id() == ptr.id => return Some(block),
```
dmonad, line 235: Do a range check whether `ptr.id` is in `block`.

yrs/src/block_store.rs
```diff
@@ -398,25 +451,39 @@ impl BlockStore {
...
         let clients = self.clients.get(&ptr.id.client)?;
-        match clients.list.get(ptr.pivot()) {
-            Some(block) if block.id().clock == ptr.id.clock => Some(block),
-            _ => {
-                // ptr.pivot missed - go slow path to find it
-                let pivot = clients.find_pivot(ptr.id.clock)?;
-                ptr.fix_pivot(pivot as u32);
-                Some(&clients.list[pivot])
+        if let Some(block) = clients.try_get(ptr.pivot()) {
+            if block.id().clock == ptr.id.clock {
```
dmonad, line 464: This should be a range check as well.

yrs/src/block_store.rs
```diff
@@ -398,25 +451,39 @@ impl BlockStore {
...
+        ptr.fix_pivot(pivot as u32);
+        clients.try_get(pivot)
+    }
+
+    /// Returns immutable reference to a block, given its pointer. Returns `None` if not such
+    /// block could be found.
+    pub(crate) fn get_block_mut(&self, ptr: &block::BlockPtr) -> Option<&mut block::Block> {
+        let clients = self.clients.get(&ptr.id.client)?;
+        if let Some(block) = clients.try_get_mut(ptr.pivot()) {
+            if block.id().clock == ptr.id.clock {
```
dmonad, line 479: This should be a range check as well.

dmonad (changes requested):

dmonad (approved):

PR #51, opened 2021-11-30
Use unicode code points indexing for text operations

Horusiath:


yrs/src/block.rs
```diff
@@ -452,7 +454,7 @@ pub(crate) struct Item {
 
     /// Used only when current item is used by map-like types. In such case this item works as a
     /// key-value entry of a map, and this field contains a key used by map.
-    pub parent_sub: Option<String>, //TODO: Rc since it's already used in Branch.map component
+    pub parent_sub: Option<Rc<str>>, //TODO: Rc since it's already used in Branch.map component
```
Horusiath, line 457: Since `parent_sub` represents a static immutable string key, that's shared between branch map component and each of the blocks that are assigned as a value to that map's entry, it's much better to share them via `Rc` rather than deep copy the strings.

yrs/src/block.rs
```diff
@@ -1039,10 +1042,7 @@ impl ItemContent {
     pub fn len(&self) -> u32 {
         match self {
             ItemContent::Deleted(deleted) => *deleted,
-            ItemContent::String(str) => {
-                // @todo this should return the length in utf16!
-                str.len() as u32
-            }
+            ItemContent::String(str) => str.chars().count() as u32,
```
Horusiath, line 1045: For sake of proper universal unicode support, we could a string length by the number of characters rather than byte length of a string.

yrs/src/block.rs
```diff
@@ -1226,6 +1226,8 @@ impl ItemContent {
                 Some(ItemContent::Any(right))
             }
             ItemContent::String(string) => {
+                // compute offset given in unicode code points into byte position
+                let offset = string.char_indices().map(|(i, _)| i).nth(offset).unwrap();
```
Horusiath, line 1230: Map by-char index onto by-byte index that's used for string splitting.

yrs/src/doc.rs
```diff
@@ -41,7 +42,7 @@ use std::cell::RefCell;
 pub struct Doc {
     /// A unique client identifier, that's also a unique identifier of current document replica.
     pub client_id: u64,
-    store: RefCell<Store>,
+    store: Rc<UnsafeCell<Store>>,
```
Horusiath, line 45: For sake of being able to pass transaction to ywasm and yffi I had to remove lifetime parameter from `Transaction<'txn>` (now it's just `Transaction`). To do so, I decided to pass store not by `RefMut` but rather by `Rc`. This allows us to have multiple concurrent transactions open at the same time. In order to be able to modify store we also needed interior mutability - I used `UnsafeCell` rather than `RefCell` as store is accessed very often and `UnsafeCell` overhead is basically 0.

yrs/src/store.rs
```diff
@@ -132,6 +133,52 @@ impl Store {
         value.clone()
     }
 
+    pub(crate) fn find_index_clean_start(&mut self, client: &u64, clock: u32) -> Option<usize> {
```
Horusiath, line 136: These two methods were just moved from `Transaction` to `Store` to make them more precise - Rust borrow checker likes precise field access.

yrs/src/transaction.rs
```diff
@@ -504,29 +479,42 @@ impl<'a> Transaction<'a> {
     /// This step is performed automatically when a transaction is about to be dropped (its life
     /// scope comes to an end).
     pub fn commit(&mut self) {
+        if self.committed {
```
Horusiath, line 482: Now transaction has a `committed` flag, which is used to prevent multiple commits over the same transaction.

yrs/src/transaction.rs
```diff
@@ -504,29 +479,42 @@ impl<'a> Transaction<'a> {
...
         self.delete_set.squash();
-        self.after_state = self.store.blocks.get_state_vector();
-
+        self.after_state = store.blocks.get_state_vector();
         // 2. emit 'beforeObserverCalls'
         // 3. for each change observed by the transaction call 'afterTransaction'
+        if !self.changed.is_empty() {
+            for (ptr, subs) in self.changed.iter() {
+                if let Some(branch) = store.get_type(ptr) {
+                    branch.trigger(self, subs.clone());
```
Horusiath, line 496: Trigger for observer API.

yrs/src/transaction.rs
```diff
@@ -578,11 +567,11 @@ impl<'a> Transaction<'a> {
         }
     }
 
-    pub(crate) fn add_changed_type(&mut self, parent: &Branch, parent_sub: Option<&String>) {
-        let trigger = match parent.item.as_ref() {
-            None => true,
-            Some(ptr) if ptr.id.clock < (self.before_state.get(&ptr.id.client)) => {
-                if let Some(item) = self.store.blocks.get_item(ptr) {
+    pub(crate) fn add_changed_type(&mut self, parent: &Branch, parent_sub: Option<Rc<str>>) {
+        let trigger = match &parent.ptr {
```
Horusiath, line 571: `Branch::item` has been removed - we can fully describe equivalent functionality using already existing `Branch::ptr` field.

yrs/src/types/mod.rs
```diff
@@ -220,14 +229,40 @@ pub struct Branch {
...
+            .field("name", &self.name)
+            .field("len", &self.len)
+            .field("type_ref", &self.type_ref)
+            .finish()
+    }
+}
+
+impl Eq for Branch {}
+
+impl PartialEq for Branch {
```
Horusiath, line 257: Since `Branch` contains event handlers now, derivation of structural equality is no longer possible.

PR #60, opened 2021-12-15
bump version to v0.2

Horusiath:


davidbrochart:
Should the `y-py` version also be bumped to 0.2 in https://github.com/yjs/y-crdt/blob/main/y-py/Cargo.toml ?

Horusiath:
We'll discuss what's necessary prior the release of y-py with @dmonad and @Waidhoferj this Friday. Once everything is set, we'll publish y-py v0.2 as well. Sorry for the confusion.

PR #61, opened 2021-12-16
Added B2, B3, B4 benches + fixed block splitting bug

Horusiath:


yrs/src/block_store.rs
```diff
@@ -562,16 +562,26 @@ impl BlockStore {
...
+
+                                blocks = self.clients.get_mut(&ptr.id.client).unwrap();
                             };
-                            let right = blocks.find(&right_ptr).unwrap();
-                            if let Some(right_item) = right.as_item_mut() {
-                                right_item.left =
-                                    Some(BlockPtr::new(right_split.id.clone(), index as u32));
-                            }
                         }
                         blocks.insert(index, Block::Item(right_split));
```
Horusiath, line 586: An old version of altered code could under some circumstances cause change of client `blocks` list to point to another client before splitted block was inserted. This could cause items created due to split with client ID `A` to be inserted into client block list of `B`.

Horusiath:
I'm attaching flamegraph results from B4 run (which takes ~72s and is ~12 slower than Yjs) - apparently majority of the time spent is in `find_pivot` method used to find blocks in the block store - which traces back to #4

PR #62, opened 2021-12-21
version upgrade: v0.2.1

Horusiath:


davidbrochart:
What about the version in https://github.com/yjs/y-crdt/blob/main/y-py/Cargo.toml ?

Horusiath:
As I couldn't run tests on my local machine to confirm if they are passing and everything is correct, I need to move to another machine before doing so. I'll try to publish it later today.

davidbrochart:
I understand, but it is important to have versions kept in sync. If you are unsure about something, then it is better not to release than to release "a part" of y-crdt.

PR #76, opened 2022-01-21
v0.4.0

Horusiath:
Finally I managed to get `cargo release` to work and released yrs packages (except y-py which still doesn't compile locally for me) - they are visible on crates.io. That's a good news. 
Bad one is: when running `cargo release minor` it handles automatically minor version upgrade and does a dry-run, which is fine. However when executing `cargo release minor --execute` in order to do actual (not dry-run) release, it bumped up the version again. Therefore I published v0.4 and skipped over v0.3.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1162,7 +1162,7 @@ dependencies = [
 
 [[package]]
 name = "ywasm"
-version = "0.3.0"
+version = "0.4.0"
 dependencies = [
  "console_error_panic_hook",
  "js-sys",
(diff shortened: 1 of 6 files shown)


dmonad:
This is nice, thank you! 

So a `cargo release minor` on the root folder performs a complete release of Yrs now? But it didn't create a tag, right?

Horusiath:
From what I've seen in the [docs](https://github.com/crate-ci/cargo-release/blob/master/docs/reference.md#config-fields) it looks like it should be capable of doing that but apparently it's not configured to do so by default.

Issue #83, opened 2022-02-04
[yffi] BUG (+fix): negative `c_char` constants are invalid on some architectures.


ngasull:
As explained by the documentation, [`c_char`](https://doc.rust-lang.org/std/os/raw/type.c_char.html) will alias `i8` or `u8` depending on the architecture. This means that negative values can't be consistently used for `c_char` constants.

Example target : `arm-linux-androideabi`

There are multiple ways to solve this, I'll suggest a PR in a second.

ngasull:
Suggested PR with details : https://github.com/y-crdt/y-crdt/pull/84

PR #84, opened 2022-02-04
[yffi] Fix negative c_char

ngasull:
Fix suggestion for https://github.com/y-crdt/y-crdt/issues/83

I also see another fix : keep using `c_char` for change the value of negative constants for something < 128.

IMO relying on i8 is just clearer as we simply are talking about constants. From the outside, it's not really important what the actual values are as the ffi user is supposed to use the constants anyway. The functions/structs expect these constants in their interface, as per their documentation, not magic values.


Horusiath (approved):
Thanks! It's much appreciated.

Horusiath:
I think, we should probably do the same for `c_long` and `c_long_long` usage and switch them to `i64` for the same reasons.

PR #87, opened 2022-02-14
[WIP] Moving Block::Skip outside of default Block impl

Horusiath:


yrs/benches/benches.rs
```diff
@@ -432,6 +432,28 @@ fn b4(c: &mut Criterion, name: &str) {
     );
 }
 
+fn b4_2(c: &mut Criterion, name: &str) {
```
Horusiath, line 435: I've added a benchmark that's used solely for checking out how long will it take to apply update generated from B4.1 (real live edit trace). Apparently applying update that's almost 400kB takes ~2-3ms.

yrs/src/block.rs
```diff
@@ -148,8 +163,7 @@ impl Block {
     pub fn last_id(&self) -> ID {
         match self {
             Block::Item(item) => item.last_id(),
-            Block::Skip(skip) => ID::new(skip.id.client, skip.id.clock + skip.len),
```
Horusiath, line 151: Since `Block::Skip` doesn't exists in block store (only used in context of update), I removed this case from default `Block` definition.

yrs/src/block.rs
```diff
@@ -503,37 +488,25 @@ pub(crate) struct Item {
 }
 
 #[derive(Debug, PartialEq, Eq, Clone)]
-pub struct Skip {
+pub struct BlockRange {
```
Horusiath, line 491: `BlockRange` covers similarities between `Skip` and `GC` blocks.

yrs/src/block.rs
```diff
@@ -573,16 +546,39 @@ impl Item {
...
             right_origin,
             content,
             parent,
             parent_sub,
-            info: info,
+            info,
+        }
+    }
+
+    pub fn try_eq(&self, other: &Item) -> Option<bool> {
```
Horusiath, line 564: I changed `Item::eq` in a way that assumes, that two items may have different neighbor `BlockPtr`'s, which are not structurally equatable, but still may point to the same blocks. Problem is that establishing equality in such way requires access to block store (in order to navigate to left/right neighbor), therefore without having it it's not always possible to do so.

yrs/src/block.rs
```diff
@@ -991,18 +998,29 @@ impl Item {
...
 
 impl SplittableString {
     pub fn len(&self, kind: OffsetKind) -> usize {
-        match kind {
-            OffsetKind::Bytes => self.content.len(),
-            OffsetKind::Utf16 => self.utf16_len(),
-            OffsetKind::Utf32 => self.unicode_len(),
+        let len = self.content.len();
+        if len == 1 {
+            len // quite often strings are single-letter, so we don't care about OffsetKind
```
Horusiath, line 1017: Micro optimization to make computing lengths for single-char strings faster (which is quite common in transaction-per-keystroke case). Problem is that it still doesn't bring enough of speedup in terms of cached `Item::len` field. We might possibly make this field to be calculated lazily.

yrs/src/block.rs
```diff
@@ -991,18 +998,29 @@ impl Item {
...
+    fn eq(&self, other: &Self) -> bool {
+        self.try_eq(other)
+            .expect(format!("Item equality was not possible: {} vs {}", self, other).as_str())
+    }
+}
+
 #[derive(PartialEq, PartialOrd, Eq, Ord, Debug, Clone)]
 pub struct SplittableString {
     content: SmallString<[u8; 8]>,
-    utf16_len: usize,
```
Horusiath, line 997: Since `Item::len` now contains precomputed field length, it doesn't have to reside here anymore.

yrs/src/update.rs
```diff
@@ -697,6 +834,117 @@ impl<T: Iterator> Memoizable for T {
     }
 }
 
+#[derive(Debug, PartialEq)]
+pub(crate) enum BlockCarrier {
+    Block(Block),
+    Skip(BlockRange),
```
Horusiath, line 840: `BlockCarrier` is essentially a wrapper around `Block` which extends it to cover `Skip` semantics for update structure purposes.

yrs/src/update.rs
```diff
@@ -60,15 +90,101 @@ impl UpdateBlocks {
...
                 }
             }
-            blocks.insert(index, Block::Item(right_split));
+            blocks.insert(index, BlockCarrier::Block(Block::Item(right_split)));
         };
     }
 }
 
+impl PartialEq for UpdateBlocks {
+    fn eq(&self, other: &Self) -> bool {
```
dmonad, line 103: We can simplify this as left/right should always be none.

dmonad (changes requested):

PR #71, opened 2022-01-20
Serde Integration for lib0::Any

GarettCooper:
Hi, I've been testing out Yrs as a CRDT solution for a primarily Rust project that I'm working on. One point of frustration which I had was the lack of integration into the greater Rust ecosystem. Since I understand that probably isn't your priority at the moment, I thought I'd contribute a first step by integrating the `lib0::Any` type with [the Serde ecosystem](https://serde.rs/#data-formats) to solve some of my own problems.

My serde implementation outperforms the hand-rolled `to_json` and `from_json` functions by a significant margin in benchmarks and is also capable of performing roundtrips in some cases where the handrolled functions switch types between `BigInt` and `Number`.

Benchmark results:
```
serde/Any-JSON roundtrip                                                                             
                        time:   [5.6044 us 5.6162 us 5.6362 us]
Found 5 outliers among 100 measurements (5.00%)
  4 (4.00%) high mild
  1 (1.00%) high severe
serde/Any serialize
                        time:   [889.08 ns 890.32 ns 892.19 ns]                                 
Found 10 outliers among 100 measurements (10.00%)
  5 (5.00%) high mild
  5 (5.00%) high severe
serde/Any deserialize   
                        time:   [891.03 ns 895.33 ns 901.11 ns]                                   
Found 15 outliers among 100 measurements (15.00%)
  2 (2.00%) low mild
  4 (4.00%) high mild
  9 (9.00%) high severe

custom json serialization/Any-JSON roundtrip                                                                             
                        time:   [21.291 us 21.438 us 21.675 us]
Found 8 outliers among 100 measurements (8.00%)
  3 (3.00%) high mild
  5 (5.00%) high severe
custom json serialization/Any serialize                                                                             
                        time:   [2.5150 us 2.5344 us 2.5716 us]
Found 5 outliers among 100 measurements (5.00%)
  1 (1.00%) high mild
  4 (4.00%) high severe
custom json serialization/Any deserialize                                                                             
                        time:   [10.502 us 10.522 us 10.550 us]
Found 7 outliers among 100 measurements (7.00%)
  3 (3.00%) high mild
  4 (4.00%) high severe
```

PR #98, opened 2022-03-30
Optional JSON serialization via serde_json

Horusiath:
Closes #71

This PR is based on @GarettCooper work. It basically includes an opt-in feature `lib0-serde`, which replaces default JSON formatter/parser with [serde_json](https://docs.serde.rs/serde_json/) lib, which is potentially faster (at the cost of transitive dependency on serde). When this feature is active, `lib0::Any` type also implements serde Serialize/Deserialize traits, so it potentially can be used by other serde-based serialization libraries.


GarettCooper (approved):
Thanks for incorporating this change! I'm trying out Yrs on its own in a completely Rust environment, which I completely understand isn't your priority, but these small integrations into the greater rust ecosystem are a a great quality of life improvement. 

As the project matures, I'd love to see more serde implementations across the different crates (for diffs and such as well). It could also be used for some great ergonomic improvements where instead of taking `Any` as a parameter, you could take `impl Serialize` instead to eliminate a lot of boilerplate for the consumer.

dmonad (approved):
Thanks @GarettCooper  & @Horusiath !

PR #105, opened 2022-04-14
Transaction Cleanup Hook

Waidhoferj:
Created an initial draft of the transaction cleanup hook. @Horusiath @dmonad I would love some feedback:
- Should I include the `after_set` or a diff between the after set and before set?
- Should any other data other than the data vectors be included?

### DIFF
--- a/yrs/src/doc.rs
+++ b/yrs/src/doc.rs
@@ -1,5 +1,5 @@
 use crate::block::ClientID;
-use crate::event::{Subscription, UpdateEvent};
+use crate::event::{EventHandler, Subscription, TransactionCleanupEvent, UpdateEvent};
 use crate::store::Store;
 use crate::transaction::Transaction;
 use crate::updates::encoder::{Encode, Encoder, EncoderV1, EncoderV2};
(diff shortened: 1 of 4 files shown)


yrs/src/transaction.rs
```diff
@@ -578,6 +578,13 @@ impl Transaction {
             }
         }
         // 8. emit 'afterTransactionCleanup'
+        if let Some(transaction_cleanup_events) = store.transaction_cleanup_events.deref() {
+            let event = TransactionCleanupEvent {
+                delete_set: self.delete_set.clone(),
```
dmonad, line 583: I suggest that we rename this `after_transaction_event` and share the `delete_set`, `after_state`, and `before_state`. This will be useful for the `UndoManager` implementation.

dmonad:
@Waidhoferj 

> Should I include the `after_state` or a diff between the after set and before set?

I think the `after_state` and `before_state` are fine as we still need them for various things (e.g. the UndoManager). 

You had an interesting idea with the `diff` approach. If we actually ended up using it, we might make some things - like calculating updates - more efficient. However, we have existing implementations that still use the `before_state` and `after_state`. In any case, we would use the diff always alongside the normal `before/after_state`. For now, we don't need it.

yrs/src/doc.rs
```diff
@@ -82,6 +82,20 @@ impl Doc {
         store.update_events.subscribe(f)
     }
 
+    pub fn on_transaction_cleanup<F>(&mut self, f: F) -> Subscription<TransactionCleanupEvent>
+    where
+        F: Fn(&Transaction, &TransactionCleanupEvent) -> () + 'static,
+    {
+        let store = unsafe { &mut *self.store.get() };
```
Horusiath, line 89: Try: `self.store_mut()`

yrs/src/doc.rs
```diff
@@ -82,6 +82,20 @@ impl Doc {
...
+        if store.transaction_cleanup_events.is_none() {
```
Horusiath, line 90: Try: `store.transaction_cleanup_events.get_or_insert_with(EventHandler::default)`

Horusiath:
@Waidhoferj could you rebase your changes on current main?

Waidhoferj:
@Horusiath just rebased.

PR #109, opened 2022-04-19
Reverse order in which update integrates client IDs

Horusiath:
Related to https://github.com/y-crdt/ypy/issues/32

### DIFF
--- a/yrs/src/update.rs
+++ b/yrs/src/update.rs
@@ -181,7 +181,7 @@ impl Update {
             let mut store = txn.store_mut();
             let mut client_block_ref_ids: Vec<ClientID> =
                 self.blocks.clients.keys().cloned().collect();
-            client_block_ref_ids.sort_by(|a, b| b.cmp(a));
+            client_block_ref_ids.sort();
 
             let mut current_client_id = client_block_ref_ids.pop();
             let mut current_target =
--- a/yrs/src/doc.rs
+++ b/yrs/src/doc.rs
@@ -347,4 +347,65 @@ mod test {
         }
         assert_eq!(txt.to_string(), "abcd".to_string());
     }
+
+    #[test]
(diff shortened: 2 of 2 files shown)


dmonad (approved):

dmonad (approved):

yrs/src/block_store.rs
```diff
@@ -173,12 +173,11 @@ impl Decode for Snapshot {
 /// A resizable list of blocks inserted by a single client.
 pub(crate) struct ClientBlockList {
     list: Vec<Box<block::Block>>,
-    integrated_len: usize,
```
Horusiath, line 176: As part of fixing an issue it turned out that `intergrated_len` is not necessary.

yrs/src/update.rs
```diff
@@ -181,7 +181,7 @@ impl Update {
             let mut store = txn.store_mut();
             let mut client_block_ref_ids: Vec<ClientID> =
                 self.blocks.clients.keys().cloned().collect();
-            client_block_ref_ids.sort_by(|a, b| b.cmp(a));
+            client_block_ref_ids.sort();
```
Horusiath, line 184: Something I've missed - a microoptimisation used in Yjs: apply blocks from highest clients to lowest in order to avoid block skipping in case of potential conflict resolution in the same-place concurrent inserts.

yrs/src/update.rs
```diff
@@ -254,20 +257,16 @@ impl Update {
                     stack_head = stack.pop();
                 } else {
                     current_target = match current_target.take() {
-                        None => None,
```
Horusiath, line 257: This is 2 of 2 bugs detected by David: in some edge cases current_target could be null. We should proceed to a next client block list.

dmonad:
Thank you!

PR #113, opened 2022-04-25
Fixes for bugs found using small data set (part 1)

Horusiath:
- applying remote delete set with client id not known on local block store
- integrate outdated YMap's block with conflict resolution combined with squashing past deleted blocks
- integrate block with removed parent

### DIFF
--- a/yrs/src/block.rs
+++ b/yrs/src/block.rs
@@ -229,8 +229,8 @@ impl BlockPtr {
                                 break;
                             }
 
-                            items_before_origin.insert(ptr.id().clone());
-                            conflicting_items.insert(ptr.id().clone());
+                            items_before_origin.insert(ptr);
+                            conflicting_items.insert(ptr);
                             if let Block::Item(item) = ptr.deref() {
                                 if this.origin == item.origin {
                                     // case 1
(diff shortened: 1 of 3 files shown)


yrs/src/block.rs
```diff
@@ -244,9 +244,13 @@ impl BlockPtr {
                                         break;
                                     }
                                 } else {
-                                    if let Some(item_origin) = item.origin {
-                                        if items_before_origin.contains(&item_origin) {
```
Horusiath, line 248: Initially we used `ID` for `items_before_origin` and `conflicting_items` while Yjs was using Item references. These are not the same (Item may be responsible for holding range of updates) as for items with squashed contents individual IDs may be not found, but the items are in there.

yrs/src/block.rs
```diff
@@ -930,12 +934,15 @@ impl Item {
...
-                if let ItemContent::Type(branch) = &item.content {
-                    self.parent = TypePtr::Branch(BranchPtr::from(branch));
+                self.parent = if let Block::Item(item) = ptr.deref() {
+                    match &item.content {
+                        ItemContent::Type(branch) => TypePtr::Branch(BranchPtr::from(branch)),
+                        ItemContent::Deleted(_) => TypePtr::Unknown,
+                        _ => panic!("Defect: parent points to a block which is not a shared type"),
+                    }
                 } else {
-                    panic!("Defect: parent points to a block which is not a shared type");
```
Horusiath, line 937: This panic could trigger i.e. if integrated item belonged to shared type that has been already removed.

PR #116, opened 2022-04-29
[WIP] Fixed issues with invalid map entry ptr setting after squashing

Horusiath:


yrs/src/update.rs
```diff
@@ -183,35 +183,37 @@ impl Update {
...
-                if local_sv.contains(id) {
+                let id = *block.id();
+                if local_sv.contains(&id) {
                     let offset = local_sv.get(&id.client) as i32 - id.clock as i32;
                     if let Some(dep) = Self::missing(&block, &local_sv) {
                         stack.push(block);
                         // get the struct reader that has the missing struct
                         match self.blocks.clients.get_mut(&dep) {
                             Some(block_refs) if !block_refs.is_empty() => {
-                                current_target = Some(block_refs);
```
Horusiath, line 205: This is an error - if we found a missing dependency block, we don't want to switch current target (which is list of updates for a current client ID), we just want to make next iteration integrate the dependent block first then continue from the last point.

This code could cause existing list of blocks to be updates for a given client ID to be dropped before they have been integrated.

yrs/src/update.rs
```diff
@@ -333,17 +333,21 @@ impl Update {
         client_block_ref_ids: &'a mut Vec<ClientID>,
         blocks: &'b mut UpdateBlocks,
     ) -> Option<(ClientID, &'b mut VecDeque<BlockCarrier>)> {
-        loop {
-            if let Some((id, Some(client_blocks))) = client_block_ref_ids
```
Horusiath, line 340: This loop would break if the poped client ID was not found on local block store. Breaking loop here means that all remaining client IDs would be dropped and never integrated.

PR #118, opened 2022-05-05
added test samples and fixes in related parts

Horusiath:
With these fixes all of the test data suites are passing. This includes `large-test-dataset` which contains over 350k test cases. However I couldn't attach it as it takes over 300MB, while the Github repo limit for file is 100MB.

It also fixes https://github.com/y-crdt/ypy/issues/45


yrs/src/block_store.rs
```diff
@@ -307,52 +305,30 @@ impl ClientBlockList {
     /// squashed into its left neighbor. In such case a squash result will be returned in order to
     /// later on rewire left/right neighbor changes that may have occurred as a result of squashing
     /// and block removal.
-    pub(crate) fn squash_left(&mut self, index: usize) -> Option<SquashResult> {
```
Horusiath, line 310: Removed squash result - now GC compaction happens immediately during squashing, which was triggering some of the errors in small-test-dataset.

yrs/src/update.rs
```diff
@@ -183,35 +183,37 @@ impl Update {
...
-                if local_sv.contains(id) {
+                let id = *block.id();
+                if local_sv.contains(&id) {
                     let offset = local_sv.get(&id.client) as i32 - id.clock as i32;
                     if let Some(dep) = Self::missing(&block, &local_sv) {
                         stack.push(block);
                         // get the struct reader that has the missing struct
                         match self.blocks.clients.get_mut(&dep) {
                             Some(block_refs) if !block_refs.is_empty() => {
-                                current_target = Some(block_refs);
```
Horusiath, line 205: If `current_target` was not empty in this branch, it would be dropped and remaining blocks would never be integrated.

yrs/src/update.rs
```diff
@@ -247,32 +249,35 @@ impl Update {
...
 
                 // iterate to next stackHead
                 if !stack.is_empty() {
                     stack_head = stack.pop();
                 } else {
-                    current_target = match current_target.take() {
-                        Some(v) if !v.is_empty() => Some(v),
+                    match current_target.take() {
+                        Some(v) if !v.is_empty() => {
+                            stack_head = v.pop_front();
```
Horusiath, line 262: There was some error here as well, but due to magic of Git cherry-pick I can no longer recall what was wrong.

yrs/src/update.rs
```diff
@@ -336,17 +337,22 @@ impl Update {
         client_block_ref_ids: &'a mut Vec<ClientID>,
         blocks: &'b mut UpdateBlocks,
     ) -> Option<(ClientID, &'b mut VecDeque<BlockCarrier>)> {
-        loop {
-            if let Some((id, Some(client_blocks))) = client_block_ref_ids
```
Horusiath, line 340: This case was too restrictive - if we returned `Some((id, None))` (meaning current block store didn't have a client id we were looking for), then all subsequent blocks from the remaining clients waiting in update were dropped.

dmonad (approved):
Nice!

PR #120, opened 2022-05-09
Added unobserve method to doc

Waidhoferj:
- Added the `unobserve_transaction_cleanup` method which mirrors the `unobserve` methods on each specific YType. This will be used to unsubscribe [the same lifecycle events in Ypy](https://github.com/y-crdt/ypy/issues/41).
- Updated the associated test case.


Horusiath:
I guess we should modify `on_transaction_cleanup` method name to `observe_transaction_cleanup` or smth - so that the naming conventions match up.

Horusiath (approved):

PR #123, opened 2022-05-12
Update yrb link

eliias:
The y-rb bindings were moved into the `y-crdt` org last week. I'll keep the GitLab mirror for CI and deployment right now, but the source of truth should be the repo here.


dmonad:
Thank you!

PR #126, opened 2022-05-19
[WIP] Doc events in ywasm & yffi

Horusiath:
1. Removed JS benchmarks from repo.
2. Renamed `Doc::on_update` &rarr; `Doc::observe_update` + added `Doc::unobserve_update`.
3. Exposed `Doc::observe_update` and `Doc::observe_after_transaction` subscriptions in Ywasm and Yffi.


yrs/src/block.rs
```diff
@@ -156,7 +156,7 @@ impl BlockPtr {
                         .blocks
                         .get_item_clean_end(&ID::new(this.id.client, this.id.clock - 1));
                     this.origin = this.left.as_deref().map(|b: &Block| b.last_id());
-                    this.content.splice(offset as usize);
+                    this.content = this.content.splice(offset as usize).unwrap();
```
Horusiath, line 159: Fixed one bug, where resending the merged block (which beginning was seen in the past) was causing a content splicing, where wrong half of spliced data was being attached.

PR #128, opened 2022-05-24
Bart/fix/utf 32

### LINKED ISSUE #124: Unicode failures
Hi, just wanted to report another issue I recently found:

Horusiath:
Fixes #124

Issue was that our block splitting method is used in two contexts:

1. Split block given user-provided index, which may use different kinds of encoding.
2. Split block given clock offset, which always (for compatibility with Yjs) uses UTF-16 encoding.

Not all places were acknowledging this distinction correctly.

Second part of an issue was related to lib0 v2 encoding: Yjs encodes multiple string as one, and then attaches an array of offsets, so that we can chunk them back. Problem was that Yjs encodes offsets using UTF-16, while Rust was using UTF-8 byte lenghts by default.


yrs/src/block.rs
```diff
@@ -713,16 +714,83 @@ impl ItemPosition {
...
-const ITEM_FLAG_COUNTABLE: u8 = 0b0010;
+const ITEM_FLAG_COUNTABLE: u8 = 0b0000_0010;
 
 /// Bit flag (1st bit) used for an item which should be kept - not used atm.
-const ITEM_FLAG_KEEP: u8 = 0b0001;
+const ITEM_FLAG_KEEP: u8 = 0b0000_0001;
+
+#[repr(transparent)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct ItemFlags(u8);
```
Horusiath, line 730: Initially I wanted to move information about string encoding onto bits of `Item::info`, but it wouldn't solve an issue (without further modifications). One good thing, that came out of it, was to change bit flags into their own dedicated type, which now is more safe and gives us explicit knowledge about which bits are used and how. It's safer and easier to debug. Also doesn't introduce any performance overhead (I've checked that).

PR #130, opened 2022-05-25
Fix: call update events callbacks on any changes in committed transactions

Horusiath:
Previous implementation on `on_update` callback was basically unchanged for over a year - it didn't do full job, as it was only called when `transaction.apply_update` was called. This PR fixes that behavior and triggers the callbacks on any changes found in the transaction during a commit. 

Changes:

- `Do::on_update` is now split into `Doc::on_update_v1` and `Doc::on_update_v2` because the nature of how these callbacks are triggered, has changes. We no longer have `Update` struct at the moment of calling them, so it's better to make it aligned with Yjs behavior.

### DIFF
--- a/yrs/src/store.rs
+++ b/yrs/src/store.rs
@@ -1,11 +1,12 @@
 use crate::block::{ClientID, ItemContent};
 use crate::block_store::{BlockStore, StateVector};
 use crate::doc::Options;
-use crate::event::{AfterTransactionEvent, EventHandler, UpdateEvent};
+use crate::event::{AfterTransactionEvent, EventHandler};
 use crate::id_set::DeleteSet;
 use crate::types::{Branch, BranchPtr, Path, PathSegment, TypeRefs};
 use crate::update::PendingUpdate;
 use crate::updates::encoder::{Encode, Encoder};
+use crate::UpdateEvent;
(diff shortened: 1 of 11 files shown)


yrs/src/store.rs
```diff
@@ -26,10 +27,6 @@ pub(crate) struct Store {
     /// operations) integrated - and therefore visible - into a current document.
     pub(crate) blocks: BlockStore,
 
-    /// Handles subscriptions for the `afterTransactionCleanup` event. Events are called with the
-    /// newest updates once they are committed and compacted.
-    pub(crate) after_transaction_events: Box<Option<EventHandler<AfterTransactionEvent>>>,
```
Horusiath, line 31: Something I've missed in @Waidhoferj PR last time - don't do `Box<Option<T>>`: it's useless. We always allocate this way. Reverse (`Option<Box<T>>`) allocates only when field is used. Additionally: `struct EventHandler(Box<HashMap>)` already boxes necessary content, so there's no need to double boxing.

dmonad (approved):

yrs/src/store.rs
Waidhoferj: Got it, thanks for the feedback!

PR #132, opened 2022-05-26
Fixed encode with offset

Horusiath:
This PR introduces several fixes that were discovered while working on `observe_udpdate_vX` observers. It turned out that transaction changed block that was merged, generated update payload would serialize entire block instead of just new part.

### DIFF
--- a/yrs/src/block.rs
+++ b/yrs/src/block.rs
@@ -498,12 +498,15 @@ impl Block {
     ) {
         match self {
             Block::Item(item) => {
+                let mut info = item.info();
                 let origin = if offset > 0 {
                     Some(ID::new(item.id.client, item.id.clock + offset - 1))
                 } else {
                     item.origin
                 };
-                let info = item.info();
+                if origin.is_some() {
+                    info |= HAS_ORIGIN;
+                }
                 let cant_copy_parent_info = info & (HAS_ORIGIN | HAS_RIGHT_ORIGIN) == 0;
                 encoder.write_info(info);
                 if let Some(origin_id) = origin {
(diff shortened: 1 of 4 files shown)


yrs/src/block.rs
```diff
@@ -498,12 +498,15 @@ impl Block {
...
             Block::Item(item) => {
+                let mut info = item.info();
                 let origin = if offset > 0 {
                     Some(ID::new(item.id.client, item.id.clock + offset - 1))
                 } else {
                     item.origin
                 };
-                let info = item.info();
+                if origin.is_some() {
+                    info |= HAS_ORIGIN;
```
Horusiath, line 508: This is the main issue - we were using native `Item::info` method. If that item didn't have left origin (eg. it was first block) `HAS_ORIGIN` flag was not set. However if we encode block with offset > 0, we were encoding that left origin, but flag was not set, causing failures during deserialization.

yrs/src/block.rs
```diff
@@ -1334,7 +1337,10 @@ impl ItemContent {
         match self {
             ItemContent::Deleted(len) => encoder.write_len(*len - offset),
             ItemContent::Binary(buf) => encoder.write_buf(buf),
-            ItemContent::String(s) => encoder.write_string(&s.as_str()[(offset as usize)..]),
+            ItemContent::String(s) => {
+                let (_, right) = s.split_at(offset as usize, OffsetKind::Utf16);
```
Horusiath, line 1341: Fix 2: offsets are clock-relative, which means they are using UTF-16 encoding. We cannot just split string content by byte offset.

yrs/src/store.rs
```diff
@@ -152,7 +152,8 @@ impl Store {
             encoder.write_uvar(clock);
             let first_block = blocks.get(start);
             // write first struct with an offset
-            first_block.encode(Some(self), encoder);
+            let offset = clock - first_block.id().clock;
+            first_block.encode_with_offset(Some(self), encoder, offset);
```
Horusiath, line 156: Fix 3: when encoding given state vector, first block must be encoded with offset, because incoming state vector clock may point inside of it.

PR #136, opened 2022-05-27
Implement public PrelimText for nesting

### LINKED ISSUE #135: Implement public PrelimText
Is there a reason why `PrelimText` is not implemented publicly? It would be nice to embed it inside a map.

https://github.com/y-crdt/y-crdt/pull/38 implemented only `PrelimArray` and `PrelimMap`. I can see that there is a `PrelimText` in `block.rs` but it returns a String type not a Text.

In my local experiments I implemented a `PrelimText2` that returns a Text type and it kind of works.

```rust
pub struct PrelimText2<'a>(pub &'a str);

impl Prelim for PrelimText2<'_> {
    fn into_content(self, _txn: &mut Transaction) -> (ItemContent, Option<Self>) {
        let inner = Branch::new(TYPE_REFS_TEXT, None);
        (ItemContent::Type(inner), Some(self))
    }

    fn integrate(self, txn: &mut Transaction, inner_ref: BranchPtr) {
        let text = Text::from(inner_ref);
        text.push(txn, self.0);
    }
}
```

I was wondering whether there are any limitations of not supporting embedding a Text type into a map.

zaynetro:
Closes #135

Let me know if this is needed or not. I will be happy to rewrite this
change if I totally got it wrong :)

### DIFF
--- a/yrs/src/block.rs
+++ b/yrs/src/block.rs
@@ -1735,9 +1735,9 @@ where
 }
 
 #[derive(Debug)]
-pub(crate) struct PrelimText(pub SmallString<[u8; 8]>);
+pub(crate) struct PrelimString(pub SmallString<[u8; 8]>);
 
-impl Prelim for PrelimText {
+impl Prelim for PrelimString {
     fn into_content(self, _txn: &mut Transaction) -> (ItemContent, Option<Self>) {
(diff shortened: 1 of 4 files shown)


Horusiath (approved):
Good job 👍 I'll include it in the next release.

PR #138, opened 2022-05-29
Remove wasm-bindgen from yrs

zaynetro:
Rand's wasm-bindgen feature will enable wasm-bindgen feature on
getrandom crate which will compile wasm-bindgen package only for wasm
target: https://github.com/rust-random/getrandom/blob/v0.1.1/Cargo.toml#L35:L36

This means that on non-wasm targets wasm-bindgen is no longer included
and reduces the amount compiled crates from 37 to 15.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -999,7 +999,6 @@ dependencies = [
  "lib0",
  "rand 0.7.3",
  "smallstr",
- "wasm-bindgen",
 ]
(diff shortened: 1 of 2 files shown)


Horusiath (approved):
Good catch. Honestly I'm surprised I missed this dependency. It's clearly unnecessary in this place.

PR #139, opened 2022-06-03
Implement Into<Any> for i64

zaynetro:
There is a BigInt type that holds an i64 but there was no `impl Into<Any>` for that.


Horusiath (approved):
Thanks 👍 I'll add it in the next release (probably some time this week).

Issue #143, opened 2022-06-18
Using lib0 in a server app


elnardu:
Hello and thanks for a great library! I am trying to use it in my app, and I noticed that for the most part lib0 assumes that its input is safe. This makes it not ideal for server applications since malicious users can easily crash your server by sending an invalid lib0 payload. 

What are your thoughts on using more `Result` types in code?

Horusiath:
Yes tbh. that was my intention originally, but it was postponed for some reasons. Maybe it's time to go back to it.

Horusiath:
We've introduced Result type for decoding (encoding doesn't use it, since atm encoders accumulate serialized data into internal `Vec<u8>` anyway). It's now available as of v0.10.

elnardu:
Oh wow! Thank you so much!

PR #144, opened 2022-06-21
lib0 type safe errors for decoding

Horusiath:
Related to #143


lib0/Cargo.toml
```diff
@@ -12,6 +12,7 @@ description = "Efficient binary encoding library for Yrs"
 lib0-serde = ["serde", "serde_json"]
 
 [dependencies]
+thiserror = "1"
```
Horusiath, line 15: Added this dependency for convenient way to create typed errors through macro definitions.

PR #148, opened 2022-07-04
Fix uninitialized memory UB in lib0

y21:
Running the lib0 decoding tests under miri reports Undefined Behavior:
```rs
Undefined Behavior: type validation failed at [0]: encountered uninitialized bytes
   --> /root/build/src/decoding.rs:171:9
    |
171 |         std::mem::transmute::<_, [u8; 8]>(b)
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ type validation failed at [0]: encountered uninitialized bytes
```
https://github.com/y-crdt/y-crdt/blob/f9018988fa8396d44b46488b24ed775a8f17aac7/lib0/src/decoding.rs#L115-L118

This transmute attempts to transforms an uninitialized `[MaybeUninit<u8>; 8]` to `[u8; 8]` which of course is illegal.

Just zero-ing the buffer leads to the exact same assembly in -O2 and -O3, so this should have no performance impact: [godbolt](https://godbolt.org/z/qfW9f9M7s)

### DIFF
--- a/lib0/src/decoding.rs
+++ b/lib0/src/decoding.rs
@@ -1,6 +1,5 @@
 use crate::error::Error;
 use crate::number::VarInt;
-use std::mem::MaybeUninit;
 
 #[derive(Default)]
 pub struct Cursor<'a> {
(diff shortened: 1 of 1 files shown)


Horusiath (approved):

Horusiath:
Thank you for the effort you made. It looks like I tried to outsmart the compiler to my own demise ;)

PR #157, opened 2022-07-10
fixed utf8/utf16 offset mapping issue

### LINKED ISSUE #156: Delete multi-byte character from text will cause panic
Hi,
I'm unsure if I misused the API. 
Should the ```len``` value pass to ```text::remove_range()``` correspond to the ```offset_kind``` of doc?

To Reproduce
```Rust
// yrs = "0.10.2"
// lib0 = "0.10.2"

use yrs::{
    types::text::*,
    Doc
};

#[test]
fn reproduce() {
    let doc = Doc::new();
    let mut txn = doc.transact();

    txt.insert(&mut txn, 0, "😭😊");
    txt.remove_range(&mut txn, 0, "😭".len() as u32);

    assert_eq!(txt.to_string().as_str(), "😊");
}
```
```
thread 'reproduce' panicked at 'byte index 2 is not a char boundary; it is inside '😭' (bytes 0..4) of `😭😊`'
```

Horusiath:
Fixes #156

### DIFF
--- a/yrs/src/types/text.rs
+++ b/yrs/src/types/text.rs
@@ -260,7 +260,9 @@ impl Text {
                                 len
                             };
                             remaining = 0;
-                            txn.store_mut().blocks.split_block(ptr, offset, encoding);
+                            txn.store_mut()
+                                .blocks
+                                .split_block(ptr, offset, OffsetKind::Utf16);
                         } else {
                             remaining -= content_len;
                         };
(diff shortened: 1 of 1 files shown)


yrs/src/types/text.rs
```diff
@@ -260,7 +260,9 @@ impl Text {
                                 len
                             };
                             remaining = 0;
-                            txn.store_mut().blocks.split_block(ptr, offset, encoding);
+                            txn.store_mut()
+                                .blocks
+                                .split_block(ptr, offset, OffsetKind::Utf16);
```
Horusiath, line 265: `offset` provided here is already mapped to the utf-16 scheme.

PR #159, opened 2022-07-12
Fixed Text::remove_range for even/odd-byte length chars

### LINKED ISSUE #156: Delete multi-byte character from text will cause panic

Horusiath:
Fixes #156 (hopefully permanently this time).

### DIFF
--- a/yrs/src/block.rs
+++ b/yrs/src/block.rs
@@ -1112,13 +1112,14 @@ impl SplittableString {
             OffsetKind::Bytes => {
                 let mut remaining = offset;
                 let mut i = 0;
-                for c in self.content.encode_utf16() {
+                // since this offset is used to splitting later on - and we can only split entire
+                // characters - we're computing by characters
+                for c in self.content.chars() {
                     if remaining == 0 {
                         break;
                     }
-                    let utf8_len = if c < 0x80 { 1 } else { 2 };
-                    remaining -= utf8_len;
-                    i += 1;
+                    remaining -= c.len_utf8() as u32;
+                    i += c.len_utf16() as u32;
                 }
                 i
             }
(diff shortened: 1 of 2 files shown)


yrs/src/block.rs
```diff
@@ -1137,12 +1138,12 @@ impl SplittableString {
...
         let mut i = 0;
-        for c in self.content.encode_utf16() {
+        for c in self.content.chars() {
             if i >= offset {
                 break;
             }
-            off += if c < 0x80 { 1 } else { 2 };
-            i += 1;
+            off += c.len_utf8() as u32;
+            i += c.len_utf16() as u32;
```
Horusiath, line 1146: Offset provided in this method is UTF-16 aligned (since its unit of measure for Yjs `item.id.clock`), while we need to convert it to UTF-8 aligned (since it's the index for Rust's string splitting).

yrs/src/block.rs
```diff
@@ -1112,13 +1112,14 @@ impl SplittableString {
...
+                // characters - we're computing by characters
+                for c in self.content.chars() {
                     if remaining == 0 {
                         break;
                     }
-                    let utf8_len = if c < 0x80 { 1 } else { 2 };
-                    remaining -= utf8_len;
-                    i += 1;
+                    remaining -= c.len_utf8() as u32;
+                    i += c.len_utf16() as u32;
```
Horusiath, line 1122: Here we're mapping offset provided by user in UTF-8 bytes (`OffsetKind::Bytes`), while what we need to return is a value provided in UTF-16 (since its unit of measure for Yjs item.id.clock).

PR #161, opened 2022-07-13
fixed invalid encoding param for block splitting in text cursor positioning

Horusiath:
Related to #156

The reason why the issue reappeared this time was due to fact, that I've assigned `OffsetKind::UTF16` correctly in text remove method but not in text find_position.

### DIFF
--- a/yrs/src/types/text.rs
+++ b/yrs/src/types/text.rs
@@ -90,7 +90,7 @@ impl Text {
                             };
                             right_ptr = store
                                 .blocks
-                                .split_block(right_ptr, offset, encoding)
+                                .split_block(right_ptr, offset, OffsetKind::Utf16)
                                 .unwrap();
                             block_len -= offset;
                             remaining = 0;
@@ -156,9 +156,13 @@ impl Text {
         mut attributes: Attrs,
     ) {
         if let Some(mut pos) = self.find_position(txn, index) {
-            pos.unset_missing(&mut attributes);
-            Text::minimize_attr_changes(&mut pos, &attributes);
-            let negated_attrs = self.insert_attributes(txn, &mut pos, attributes);
+            let negated_attrs = if !attributes.is_empty() {
+                pos.unset_missing(&mut attributes);
+                Text::minimize_attr_changes(&mut pos, &attributes);
+                Some(self.insert_attributes(txn, &mut pos, attributes))
+            } else {
+                None
+            };
 
             let value = crate::block::PrelimString(chunk.into());
             let item = txn.create_item(&pos, value, None);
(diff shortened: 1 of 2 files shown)


yrs/src/types/text.rs
```diff
@@ -90,7 +90,7 @@ impl Text {
                             };
                             right_ptr = store
                                 .blocks
-                                .split_block(right_ptr, offset, encoding)
+                                .split_block(right_ptr, offset, OffsetKind::Utf16)
```
Horusiath, line 93: This was an source of the bugs.

yrs/src/types/text.rs
```diff
@@ -156,17 +156,23 @@ impl Text {
...
-            pos.unset_missing(&mut attributes);
-            Text::minimize_attr_changes(&mut pos, &attributes);
-            let negated_attrs = self.insert_attributes(txn, &mut pos, attributes);
+            let negated_attrs = if !attributes.is_empty() {
+                pos.unset_missing(&mut attributes);
+                Text::minimize_attr_changes(&mut pos, &attributes);
+                Some(self.insert_attributes(txn, &mut pos, attributes))
+            } else {
+                None
+            };
```
ArsnealX, line 165: I want to raise a question I have about this code here. It seems to cause it to be hard to insert a string with no attributes into one with attributes.

ArsnealX: I'm not sure if this is something that should not be allowed, but the ```Diff``` struct defines ```Attrs``` as optional. So I think it might be worth a mention.
https://github.com/y-crdt/y-crdt/blob/96aa3d1b24754400fbd9013235e5edbd97fd22f0/yrs/src/types/text.rs#L695-L697

ArsnealX: Add a test:
```Rust
use yrs::{
    types::{text::Diff, Attrs},
    Doc,
};

#[test]
fn insert_string_with_no_attribute() {
    let doc = Doc::new();
    let mut txn = doc.transact();

    let attrs = Attrs::from([("a".into(), "a".into())]);
    txt.insert_with_attributes(&mut txn, 0, "ac", attrs.clone());
    txt.insert_with_attributes(&mut txn, 1, "b", Attrs::new());

    let expect = vec![
        Diff::Insert("a".into(), Some(Box::new(attrs.clone()))),
        Diff::Insert("b".into(), None),
        Diff::Insert("c".into(), Some(Box::new(attrs.clone()))),
    ];

    assert!(txt.diff(&mut txn).eq(&expect))
}
```

Horusiath: My bad here. I thought that this was a micro optimisation that doesn't interfere with library behaviour - hence no tests were broken. I put too much trust into our test suite. Thank you for noticing that.

PR #168, opened 2022-07-21
Optimize find_position performance

ArsnealX:
Hi,
Recently, I was preparing to move to yrs from yjs on my iOS app due to unstable WebKit and relevant slow JavaScriptCore. So I ran some performance tests. One of those tests contains 12000+ inserts with all kinds of different attributes and takes approximately 70 seconds to finish on an iPhone 6s (40 seconds on i9-9880H iOS Simulator), while yjs on the same device using JavaScriptCore only take around a second.

Down below are some profiling info from the Instruments app of Xcode (Weight and Symbol).
```
1.17 min 93.8%   yrs::types::text::Text::insert_with_attributes
1.17 min 93.5%   yrs::types::text::Text::find_position
57.89 s  77.0%   yrs::types::text::Text::update_current_attributes
21.84 s  29.0%   _$LT$hashbrown..raw..RawTable$LT$T$C$A$GT$$u20$as$u20$core..clone..Clone$GT$::clone
9.29 s   12.3%   _$LT$hashbrown..raw..RawTable$LT$T$C$A$GT$$u20$as$u20$core..clone..Clone$GT$::clone
```

This pull request optimizes partial performance problems with ```find_position()```. It maintains a hashmap of attribute key to ```BlockPtr``` and clones the value in ```BlockPtr``` to ```current_attrs``` after the loop end instead of cloning the attribute value every time in the loop. The execution time of the case mentioned above improved to 11 seconds on the iPhone 6s (2.5 seconds on i9-9880H iOS Simulator). 

Here is the benchmark function I used to test the code:
```Rust
use criterion::*;
use yrs::Doc;
use yrs::types::Attrs;
use lib0::any::Any;
use std::rc::Rc;

fn benchmark(c: &mut Criterion) {
    let mut group = c.benchmark_group("group");
    group.significance_level(0.1).sample_size(10);
    group.bench_function("Insert text with attributes", |b| {
        b.iter(|| {
            let doc = Doc::new();
            let mut txn = doc.transact();
            let txt = txn.get_text("text");
            for i in 0..12000 {
                txt.insert_with_attributes(&mut txn, i, "a", Attrs::from(
                    [
                        (Rc::from("a".to_string()), Any::String(Box::from(i.to_string()))),
                    ]
                ));
            }
        });
    });
    group.finish();
}

criterion_group!(benches, benchmark);
criterion_main!(benches);
```
Before this pull request: (i9-9880H)
```
group/Insert text with attributes                                                                          
                        time:   [13.607 s 13.825 s 14.072 s]
```
After: (i9-9880H)
```
group/Insert text with attributes                                                                          
                        time:   [2.1896 s 2.2473 s 2.3098 s]
                        change: [-84.254% -83.744% -83.240%] (p = 0.00 < 0.10)
                        Performance has improved.
```
A test using yjs for the reference:
```JavaScript
function benchmark() {
    let start = performance.now()
    const doc = new Y.Doc()
    const ytext = doc.getText('text')
    doc.transact(function () {
        for (let i = 0; i < 12000; i++) {
            ytext.insert(i, "a", { "a": i.toString() })
        }
    })
    
    console.log('benchmark insert with attribute: ', performance.now() - start)
}
```
Result: (i9-9880H, Node.js v16.15.1)
```
benchmark insert with attribute: 1736.1995700001717
```

By the way, I found that the search marker feature is not in yrs yet. I hope that will sort out the rest of the performance issues. I would be so grateful if you let me know if that is in the current development plan.

### DIFF
--- a/yrs/src/types/text.rs
+++ b/yrs/src/types/text.rs
@@ -63,6 +63,7 @@ impl Text {
             }
         };
 
+        let mut format_ptrs = HashMap::new();
         let store = txn.store_mut();
         let encoding = store.options.offset_kind;
         let mut remaining = index;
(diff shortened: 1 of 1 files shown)


Horusiath:
Thanks @ArsnealX - we didn't work on optimising find_position atm. simply because for client apps, insert operations are bound to user reaction time: it's highly doubtful that any human being is capable to making 12000 key strokes in 70 secs.

Eventually, we'll planning on implementing 2 ways to improve the speed of insert ops:

1. Implement and expose cursors API (it's more advanced idea of search markers, conceptually similar to how cursors in text editors work), which are structure that refers to an actual position in the text/array and can advance together with user insert/delete operations.
2. Reimplement `ClientBlockList` as an rope data structure, which has logarithmic random lookup complexity (current impl. has linear).

ArsnealX:
@Horusiath Thank you for keeping me in the loop. I'm looking forward to these changes coming soon. 

The benchmarks are coming from our user case in the production app:

 - When we init an existing rich-text document to ydoc, we need to split the text with different attributes to multiply insert operations for future editing.

- The user can simply copy and paste a huge text with complex attributes into the document. 

- Some of our users have a large to-do (approx 10000+ lines) in a document, the larger the doc, the worse the performance.

Maybe there are alternative ways to bypass the massive insert operations to create a new ydoc, we're looking forward to your suggestions.

Horusiath (approved):
👍 It looks fine. Thanks for thorough implementation and benchmarks.

PR #169, opened 2022-07-22
The scope of the unsafe block can be appropriately reduced

cactter:
In this function you use the unsafe keyword for some safe expressions.

We need to mark unsafe operations more precisely using unsafe keyword. Keeping unsafe blocks small can bring many benefits. For example, when mistakes happen, we can locate any errors related to memory safety within an unsafe block. This is the balance between Safe and Unsafe Rust. The separation is designed to make using Safe Rust as ergonomic as possible, but requires extra effort and care when writing Unsafe Rust.

Hope this PR can help you.
Best regards.
**References**
https://doc.rust-lang.org/nomicon/safe-unsafe-meaning.html
https://doc.rust-lang.org/book/ch19-01-unsafe-rust.html

### DIFF
--- a/yrs/src/block.rs
+++ b/yrs/src/block.rs
@@ -350,11 +350,11 @@ impl BlockPtr {
                 item.content.gc();
                 let len = item.len();
                 if parent_gced {
-                    unsafe {
+                    
                         let gc = Block::GC(BlockRange::new(item.id, len));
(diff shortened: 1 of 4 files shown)


Horusiath (approved):
👍 I don't mind these changes. While immediate benefits may not be clearly visible, this indeed may help us track some buggy behaviors in the future.

PR #171, opened 2022-07-23
Implement a Serializer and Deserializer for lib0::Any

GarettCooper:
With a confusingly similar name to my previous PR, this PR implements a Serializer and Deserializer for `lib0::Any`, allowing any Rust type which implements the `Serialize`/`Deserialize` trait to be converted directly into instances of `Any`.

The use case I'm experimenting with Yrs for requires putting Rust objects of many different types into a `Map`, so I decided to go ahead and implement this change when I got sick of manually implementing `Into<Any>`/`TryFrom<Any>` for each of them.

```rust
map.insert(transaction, id.to_string(), item.into());

impl Into<Any> for Item {
    fn into(self) -> Any {
        Any::Map(Box::new(HashMap::from([
            ("field_name".to_string(), self.field_name.into()),
            ("field_name".to_string(), self.field_name.into()),
            ("field_name".to_string(), Any::BigInt(self.field_name as i64)),
            ("field_name".to_string(), self.field_name.into()),
        ])))
    }
}
```

Becomes:

```rust
map.insert(transaction, id.to_string(), serialize_to_any(&item).unwrap());
```

This functionality is still behind the `lib0-serde` feature, so it won't burden any of the targets which aren't making use of it.

I'm not particularly happy with the names of the `serialize_to_any` and `deserialize_from_any` functions. I though about copying the serde_json naming scheme, but their functions are named with the expectation of carrying the `serde_json` context with them (like `serde_json::from_str()`), which doesn't carry over very neatly to lib0. I also considered `to_any` and `from_any`, but that might create confusion with the existing `Into<Any>` implementations. Any alternative suggestions are appreciated.


Horusiath:
I don't mind `to_any`\`from_any` naming convention. Similar convention already exists in some places in Rust (like `to_string`->`Into<String>` and `to_vec`->`Into<Vec<T>>`).

lib0/src/serde/de.rs
```diff
@@ -0,0 +1,905 @@
...
+use serde::de::value::{MapAccessDeserializer, MapDeserializer, SeqDeserializer};
+use serde::de::{IntoDeserializer, MapAccess, SeqAccess, Visitor};
+use serde::{Deserialize, Deserializer};
+use std::any::type_name;
+use std::collections::HashMap;
+use std::convert::TryInto;
+use std::fmt::{Display, Formatter};
+use thiserror::Error;
+
+pub fn from_any<'de, T: Deserialize<'de>>(any: &'de Any) -> Result<T, AnyDeserializeError> {
```
Horusiath, line 11: I think that `from_any`/`to_any` could be put behind the traits and be implemented by default for any `T` that implements serde Deserialize.

Horusiath (approved):
Thanks for your contribution. I'm sorry it took so long to review this. It looks great, and I can address my remarks by myself later.

IndexXuan:
why all the int became BigInt. which make yjs root.toJSON() with something like
```json
{
  "width": 1440n
}
```

Issue #188, opened 2022-09-13
Error to decode update v2 on specific case


namse:
yrs version: 0.11.2

I failed to decode update_v2 with this buffer,

```rust
#[test]
fn yrs_decode_test() {
    use yrs::updates::decoder::Decode;

    let buffer = [
        0x00, 0x00, 0x06, 0xd0, 0x87, 0xf7, 0xf9, 0x0d, 0x02, 0x03, 0x02, 0x01, 0x00, 0x00, 0x07,
        0x28, 0x00, 0x27, 0x00, 0x28, 0x01, 0x27, 0x52, 0x47, 0x72, 0x6f, 0x6f, 0x74, 0x5f, 0x5f,
        0x76, 0x65, 0x72, 0x73, 0x69, 0x6f, 0x6e, 0x5f, 0x5f, 0x72, 0x6f, 0x6f, 0x74, 0x73, 0x65,
        0x71, 0x75, 0x65, 0x6e, 0x63, 0x65, 0x69, 0x64, 0x6e, 0x61, 0x6d, 0x65, 0x63, 0x75, 0x74,
        0x73, 0x72, 0x6f, 0x6f, 0x74, 0x63, 0x68, 0x61, 0x72, 0x61, 0x63, 0x74, 0x65, 0x72, 0x73,
        0x72, 0x6f, 0x6f, 0x74, 0x66, 0x61, 0x63, 0x65, 0x5f, 0x65, 0x78, 0x70, 0x72, 0x65, 0x73,
        0x73, 0x69, 0x6f, 0x6e, 0x73, 0x04, 0x0b, 0x04, 0x08, 0x02, 0x44, 0x01, 0x0a, 0x04, 0x10,
        0x05, 0x01, 0x01, 0x00, 0x02, 0x01, 0x03, 0x01, 0x00, 0x01, 0x02, 0x41, 0x01, 0x01, 0x07,
        0x00, 0x7d, 0x01, 0x77, 0x15, 0x56, 0x39, 0x55, 0x6b, 0x39, 0x70, 0x78, 0x55, 0x4b, 0x5a,
        0x49, 0x72, 0x57, 0x36, 0x63, 0x4f, 0x6b, 0x43, 0x30, 0x52, 0x67, 0x77, 0x0c, 0x6e, 0x65,
        0x77, 0x20, 0x73, 0x65, 0x71, 0x75, 0x65, 0x6e, 0x63, 0x65, 0x00,
    ];

[...]

PR #189, opened 2022-09-13
Decode UintOptRle with floating number for negative zero

namse:
Related issue: https://github.com/y-crdt/y-crdt/issues/188

I found that lib0 has wrong port about negative value checking for negative zero in integer type. 
I changed it to use floating number for `UIntOptRleDecoder`.

I'm not sure that we have to cover all cases or just `UIntOptRleDecoder`.


lib0/src/number.rs
```diff
@@ -284,3 +284,67 @@ fn read_var_i64<R: Read>(reader: &mut R) -> Result<i64, Error> {
         }
     }
 }
+
+impl VarInt for f64 {
```
namse, line 288: It would sounds strange that Var`Int` but impl for `f`64.

Horusiath:
First thank you for your investigation. However I don't think it's a good idea to indtroduce `VarInt` trait for `f64`, simply because the issue is not about floating point numbers. The source of an issue exists because `VarInt::read` doesn't include sign information in case of `-0`. It seems that better idea would be to have a dedicated method like `fn read_var_with_sign() -> Result<(i64,i8), Error>` which will return sign information together with a decoded value.

namse:
Sure, I think so. here is my suggestion.

```rust
struct Signed<T> {
  value: T,
  sign: bool,
}

impl<T> Signed<T> {
  pub fn is_positive(&self) -> bool {
    self.sign
  }
  pub fn is_negative(&self) -> bool {
    -self.sign
  }
}

fn read_var_signed() -> Result<Signed, Error> {
  ...
}
```

@Horusiath 
I haven't checked all the code but just made this PR, so I'm a little worried that if I do this it will be insufficient work.
Will you do the sign work yourself? If so, I'll close the PR.
Please let me know if you think it would be better for me to do it.

Horusiath:
I like you suggestion. It expresses the intent in clean way. Feel free to contribute with it. Otherwise I'll probably work on it on the weekend/next week.

Issue #195, opened 2022-09-28
Thread-safe API


Horusiath:
# Motivation

Unlike Yjs, Yrs can be used in situations, when document contents are accessed concurrently from mulitple threads. Right now we don't provide API that supports rust-idomatic safety features at a type level. We're in a process of working on it atm.

We **don't** aim to give a full thread safety features (like internal use of locks and mutexes), as they may differ depending on runtime in use (eg. OS threads vs. async code) and compilation target (eg. Web Assembly). Instead we want to get the most out of Rust `Send`/`Sync` capable types at compile time and supplement them with internal mutability in (borrow/borrow_mut) in places where compile time support is not possible.

# How

Our strategy towards thread safe `Doc` access is oriented around splitting transactions between shared read-only ones and exclusive read-write transactions. Some initial thoughts on that were proposed some time ago in #69 . **tl;dr** we're going to offer two kinds of transactions:

```rust
pub struct Doc {
  store: Rc<RefCell<Store>>
}

// the same API will be introduced to BranchPtr
impl Doc {
  // try to obtain transaction lock in shared mode
  pub fn try_transact(&self) -> Result<Transaction, BorrowError> {
    let store: Ref<_> = self.store.try_borrow()?;
    Ok(Transaction(store))
  }

  // try to obtain transaction lock in exclusive mode
  pub fn try_transact_mut(&self) -> Result<TransactionMut, BorrowMutError>{
    let store: RefMut<_> = self.store.try_borrow_mut()?;
    Ok(TransactionMut::new(store))
  }

  pub fn transact(&self) -> Transaction {
    self.try_transact().expect("another read-write transaction is in progress")
  }

  pub fn transact_mut(&self) -> TransactionMut {
    self.try_transact_mut().expect("another transaction is in progress")
  }
}

// read-only transaction is just super lightweight wrapper around the store reference
pub struct Transaction<'doc>(Ref<'doc,Store>);

impl<'doc> ReadTxn for Transaction<'doc> { /* todo */ }

pub struct TransactionMut<'doc> {
  store: RefMut<'doc, Store>,
  // ... other fields
}

// TransactionMut can be used by all methods (both using readonly access to document store 
// and mutating it)
impl<'doc> ReadTxn for TransactionMut<'doc> { /* todo */ }
impl<'doc> WriteTxn for TransactionMut<'doc> { /* todo */ }

pub trait ReadTxn {
  fn store(&self) -> &Store;
}

pub trait WriteTxn {
[...]

PR #199, opened 2022-10-26
Multithreading support

Horusiath:
Related: #195


tests-wasm/y-array.tests.js
```diff
@@ -11,8 +11,8 @@ export const testInserts = tc => {
     t.compare(d1.id, 1)
     var x = d1.getArray('test');
 
-    d1.transact(txn => x.insert(txn, 0, [1, 2.5, 'hello', ['world'], true]))
-    d1.transact(txn => x.push(txn, [{key:'value'}]))
+    x.insert(0, [1, 2.5, 'hello', ['world'], true])
```
Horusiath, line 14: Nice result of having implicit transactions is that we no longer need to build transaction scope by hand.

tests-wasm/y-map.tests.js
```diff
@@ -124,8 +126,8 @@ export const testObserver = tc => {
 
     // remove an entry and update another on
     d1.transact(txn => {
-        x.delete(txn, 'key1')
-        x.set(txn, 'key2', 'value2')
+        x.delete('key1', txn)
```
Horusiath, line 129: Transactions can now be also provided explicitly as the last parameter.

yrs/src/block.rs
```diff
@@ -523,114 +527,93 @@ impl PartialEq for BlockPtr {
...
-#[derive(PartialEq)]
-pub(crate) enum Block {
-    /// An active block containing user data.
-    Item(Item),
-
-    /// Block, which is a subject of garbage collection after an [Item] has been deleted and its
-    /// safe for the transaction to remove it.
-    GC(BlockRange),
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub(crate) struct BlockSlice {
```
Horusiath, line 531: Block slice represents a range of values within a single Block, without need of slicing that block.

yrs/src/block.rs
```diff
@@ -1387,45 +1433,88 @@ impl ItemContent {
...
-            ItemContent::String(v) => v
-                .chars()
-                .map(|c| Value::Any(Any::String(c.to_string().into_boxed_str())))
-                .collect(),
-            ItemContent::Type(c) => {
-                let branch_ref = BranchPtr::from(c);
-                vec![branch_ref.into()]
+    /// Reads a contents of current [ItemContent] into a given `buf`, starting from provided
+    /// `offset`. Returns a number of elements read this way (it cannot be longer than `buf`'s len.
+    pub fn read(&self, offset: usize, buf: &mut [Value]) -> usize {
```
Horusiath, line 1438: A new way of reading the contents of the `ItemContent` - previously we had `get_content`, `content_first` and `content_last` to read the all block values or the first/last value. With this we can provide a buffer of fixed length and read block values at any given offset we like.

yrs/src/block_store.rs
```diff
@@ -425,29 +425,25 @@ impl BlockStore {
...
+        let blocks = self.clients.get(&id.client)?;
+        let index = blocks.find_pivot(id.clock)?;
+        let ptr = blocks.get(index);
+        let offset = id.clock - ptr.id().clock;
+        Some(BlockSlice::new(ptr, offset, ptr.len() - 1))
     }
 
-    pub(crate) fn get_item_clean_end(&mut self, id: &ID) -> Option<BlockPtr> {
-        let blocks = self.clients.get_mut(&id.client)?;
+    pub(crate) fn get_item_clean_end(&self, id: &ID) -> Option<BlockSlice> {
```
Horusiath, line 436: `get_item_clean_start` and `get_item_clean_end` now are returning `BlockSlice` which means, that they don't split blocks right away. Store has a dedicated method that can split block along the slice boundaries when necessary.

yrs/src/doc.rs
```diff
@@ -67,10 +71,102 @@ impl Doc {
...
+    /// collaborative text editing: they expose operations to append and remove chunks of text,
+    /// which are free to execute concurrently by multiple peers over remote boundaries.
+    ///
+    /// If not structure under defined `name` existed before, it will be created and returned
+    /// instead.
+    ///
+    /// If a structure under defined `name` already existed, but its type was different it will be
+    /// reinterpreted as a text (in such case a sequence component of complex data type will be
+    /// interpreted as a list of text chunks).
+    pub fn get_text(&self, name: &str) -> Text {
```
Horusiath, line 84: Now root-level type acquisition happens straight at the document level (without need of transaction). The reason is that building a new y-type requires passing a ref counted pointer to a Store (for purposes of transaction creation), which is available only at the document level or at another y-type level - transaction has only temporary reference to a store, but not a ref counted reference.

yrs/src/doc.rs
```diff
@@ -209,13 +293,75 @@ pub enum OffsetKind {
...
+
+    fn try_transact_mut(&self) -> Result<TransactionMut, BorrowMutError> {
+        let store = self.store.as_ref().unwrap();
+        Ok(TransactionMut::new(store.try_borrow_mut()?))
+    }
+}
+
+impl<T> Transact for T
+where
+    T: AsRef<Branch>,
```
Horusiath, line 343: Now a new transaction can be created at any point in time either from the `Doc` level or straight from the y-types.

yrs/src/observer.rs
```diff
@@ -0,0 +1,274 @@
+use crate::atomic::AtomicRef;
+use std::fmt::{Debug, Formatter};
+use std::sync::atomic::{AtomicU32, Ordering};
+use std::sync::Arc;
+
+pub type SubscriptionId = u32;
+
+#[derive(Debug)]
+pub struct Observer<F: Clone> {
```
Horusiath, line 9: Observer replaces an old `EventHandler`. It's thread safe without locks and it's using atomic hardware intrinsic ops combined with copy-on-write pattern to perform updates.

yrs/src/store.rs
```diff
@@ -254,6 +258,46 @@ impl Store {
             None
         }
     }
+
+    /// Consumes current block slice view, materializing it into actual memory layout,
+    /// splitting underlying block along [BlockSlice::start]/[BlockSlice::end] offsets.
+    ///
+    /// Returns a block created this way, that represents the boundaries that current [BlockSlice]
+    /// was representing.
+    pub(crate) fn materialize(&mut self, mut slice: BlockSlice) -> BlockPtr {
```
Horusiath, line 267: This method takes a block slice and slices the underlying block accordingly to a ranges defined by that slice.

yrs/src/transaction.rs
```diff
@@ -92,109 +44,120 @@ impl Transaction {
...
-    /// If a structure under defined `name` already existed, but its type was different it will be
-    /// reinterpreted as a text (in such case a sequence component of complex data type will be
-    /// interpreted as a list of text chunks).
-    pub fn get_text(&mut self, name: &str) -> Text {
-        let mut c = self
-            .store_mut()
-            .get_or_create_type(name, None, TYPE_REFS_TEXT);
-        c.store = Some(self.store.clone());
-        Text::from(c)
+    fn encode_state_as_update<E: Encoder>(&self, sv: &StateVector, encoder: &mut E) {
```
Horusiath, line 57: These methods are not defined straight on `ReadTxn` which is a trait, so `pub` encapsulation modifier is no longer needed.

yrs/src/transaction.rs
```diff
@@ -92,109 +44,120 @@ impl Transaction {
...
-        );
-        c.store = Some(self.store.clone());
-        XmlElement::from(c)
+pub trait WriteTxn: Sized {
+    fn store_mut(&mut self) -> &mut Store;
+}
+
+#[derive(Debug)]
+pub struct Transaction<'doc> {
+    store: Ref<'doc, Store>,
```
Horusiath, line 83: Read-only transaction is just a temporary readonly reference to a block store.

tests-wasm/y-map.tests.js
```diff
@@ -76,21 +76,23 @@ export const testIterator = tc => {
...
         'b': 2,
         'c': 3
     }
-    for (let [key, value] of x.entries()) {
-        let v = expected[key]
-        t.compare(value, v)
-        delete expected[key]
-    }
+    d1.transact(txn => {
+        for (let [key, value] of x.entries(txn)) {
```
dmonad, line 90: Just a note: we can probably make `txn` optional in `.entries` as well.

yrs/src/doc.rs
```diff
@@ -209,13 +293,75 @@ pub enum OffsetKind {
...
+
+    /// Creates a transaction used for all kind of block store operations.
+    /// Transaction cleanups & calling event handles happen when the transaction struct is dropped.
+    fn transact(&self) -> Transaction {
+        self.try_transact().expect("cannot read document store contents because another read-write transaction is in progress")
+    }
+
+    /// Creates a transaction used for all kind of block store operations.
+    /// Transaction cleanups & calling event handles happen when the transaction struct is dropped.
+    fn transact_mut(&self) -> TransactionMut {
```
dmonad, line 313: From the docs the difference between `transact` and `transact_mut` is not clear.

The `*_mut` part might be confusing, because both transaction types return an object. Something like `transact_rw` & `transact_ro` might be better.

dmonad (commented):
How does this affect the benchmarks?

tests-wasm/y-map.tests.js
Horusiath: I was talking about this yesterday. The issue is that:

1. wasm_bindgen requires structs to be non-generic.
2. optional transactions are passed using reference `&Transaction`, but if it was not provided we need to instantiate them as non-reference `Transaction`
3. it means that for every iterator we need to have two types of structs defined: one for reference transaction and one for owned transaction.

This causes a lot of boilerplate to make. I can think about what to do with that.

yrs/src/doc.rs
Horusiath: This is something that I've seen in other rust libraries. `mut` suffix is a standard for methods returning objects that can be mutated or used to mutate data structures (eg. `RefCell::borrow` returns `Ref` which is readonly reference, while `RefCell::borrow_mut` returns `RefMut` which is mutable reference). For this reason I decided to keep that naming convention in yrs.

chanced:
Seconded that `_mut` is understood in the rust ecosystem.

yrs/src/doc.rs
chanced: Seconded that _mut is understood in the rust ecosystem.

tests-wasm/y-map.tests.js
Horusiath: After few days of fighting with borrow checker (see commit: https://github.com/y-crdt/y-crdt/pull/199/commits/904839fd6cf3721287d116a30d916e0351aad56a) to make implicit transactions possible in this context, I realized that doing iterators in ywasm may not be possible - the reason is that every WASM object must be deallocated explicitly by calling `ywasmValue.free()` in JavaScript. If we don't do that iterator will hold the transaction forever, making other operations impossible.

Initially I wanted to use iterator [return](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#returnvalue) method to do the transaction cleanup. But there's a problem - it's only called when loop exits prematurelly (eg. due to `break;`). If loop ends normally, `iterator.return()` is never called, and since iterator is created by the loop itself under the hood, we also don't have reference that we could call free on.

I'm going to rewrite ywasm iterators to ordinary JS arrays to avoid memory leaks. Question is: **Should we leave changes in yrs iterators from commit https://github.com/y-crdt/y-crdt/pull/199/commits/904839fd6cf3721287d116a30d916e0351aad56a? They allow iterators to take ownership over passed transaction or just use transaction reference, which means that implicit iterators could potentially be used in other languages (y-py, y-rb), that can collect garbage after themselves**

/cc @dmonad @Waidhoferj @eliias

Horusiath:
I'm adding benchmarks compared to current main:

```
[B1.1] Append N characters/6000
                        time:   [15.749 ms 16.307 ms 16.524 ms]
                        change: [-9.9227% -1.7960% +6.5006%] (p = 0.70 > 0.05)
                        No change in performance detected.

[B1.2] Insert string of length N/1
                        time:   [6.5101 us 6.5198 us 6.5274 us]
                        change: [+1.2870% +1.9746% +2.5810%] (p = 0.00 < 0.05)
                        Performance has regressed.

[B1.3] Prepend N characters/6000
                        time:   [4.1794 ms 4.1900 ms 4.2093 ms]
                        change: [+0.5192% +1.1384% +1.7386%] (p = 0.00 < 0.05)
                        Change within noise threshold.

[B1.4] Insert N characters at random positions/6000
                        time:   [67.853 ms 67.895 ms 67.956 ms]
                        change: [+0.8957% +1.0941% +1.2926%] (p = 0.00 < 0.05)
                        Change within noise threshold.

[B1.5] Insert N words at random positions/6000
                        time:   [21.511 ms 21.686 ms 22.024 ms]
                        change: [+1.6818% +3.0298% +4.4452%] (p = 0.00 < 0.05)
                        Performance has regressed.

Benchmarking [B1.7] Insert/Delete strings at random positions/6000: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 6.7s or enable flat sampling.
[B1.7] Insert/Delete strings at random positions/6000
                        time:   [277.66 ms 335.69 ms 369.75 ms]
                        change: [-20.365% -2.3014% +20.249%] (p = 0.83 > 0.05)
                        No change in performance detected.

[B1.8] Append N numbers/6000
                        time:   [5.3259 ms 5.3355 ms 5.3414 ms]
                        change: [+1.2937% +1.7541% +2.2328%] (p = 0.00 < 0.05)
                        Performance has regressed.

[B1.9] Insert Array of N numbers/1
                        time:   [83.859 us 262.52 us 406.38 us]
                        change: [-54.710% +0.5965% +116.52%] (p = 0.99 > 0.05)
                        No change in performance detected.

[B1.10] Prepend N numbers/6000
                        time:   [13.787 ms 15.793 ms 19.117 ms]
                        change: [-2.9393% +15.368% +37.667%] (p = 0.16 > 0.05)
                        No change in performance detected.

Benchmarking [B1.11] Insert N numbers at random positions/6000: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 5.9s or enable flat sampling.
[B1.11] Insert N numbers at random positions/6000
                        time:   [107.63 ms 108.30 ms 109.30 ms]
                        change: [+35.070% +36.203% +37.444%] (p = 0.00 < 0.05)
                        Performance has regressed.

[B2.1] Concurrently insert string of length N at index 0/6000
                        time:   [52.222 us 52.511 us 52.838 us]
                        change: [+28.196% +29.300% +30.410%] (p = 0.00 < 0.05)
                        Performance has regressed.

[B2.2] Concurrently insert N characters at random positions/6000
                        time:   [218.90 ms 219.60 ms 220.17 ms]
                        change: [-5.0611% -2.7845% -0.5921%] (p = 0.03 < 0.05)
                        Change within noise threshold.

[B2.3] Concurrently insert N words at random positions/6000
                        time:   [395.14 ms 396.02 ms 396.93 ms]
                        change: [-2.7521% -1.5558% -0.6782%] (p = 0.01 < 0.05)
                        Change within noise threshold.

[B2.4] Concurrently insert & delete/6000
                        time:   [489.74 ms 536.33 ms 585.08 ms]
                        change: [-13.253% -2.0288% +11.536%] (p = 0.76 > 0.05)
                        No change in performance detected.

Benchmarking [B3.1] 20√N clients concurrently set number in Map/1540: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 9.8s or enable flat sampling.
[B3.1] 20√N clients concurrently set number in Map/1540
                        time:   [174.22 ms 175.45 ms 177.18 ms]
                        change: [+2.5833% +4.0492% +5.3495%] (p = 0.00 < 0.05)
                        Performance has regressed.

Benchmarking [B3.2] 20√N clients concurrently set Object in Map/1540: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 9.7s or enable flat sampling.
[B3.2] 20√N clients concurrently set Object in Map/1540
                        time:   [171.89 ms 174.55 ms 178.33 ms]
                        change: [-2.8959% -1.8191% -0.5386%] (p = 0.01 < 0.05)
                        Change within noise threshold.

Benchmarking [B3.3] 20√N clients concurrently set String in Map/1540: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 9.6s or enable flat sampling.
[B3.3] 20√N clients concurrently set String in Map/1540
                        time:   [175.70 ms 175.89 ms 176.27 ms]
                        change: [+8.1823% +8.5839% +8.9913%] (p = 0.00 < 0.05)
                        Performance has regressed.

Benchmarking [B3.4] 20√N clients concurrently insert text in Array/1540: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 9.4s or enable flat sampling.
[B3.4] 20√N clients concurrently insert text in Array/1540
                        time:   [172.67 ms 174.38 ms 177.07 ms]
                        change: [+0.1421% +1.0195% +1.7867%] (p = 0.03 < 0.05)
                        Change within noise threshold.

[B4.2] Apply real-world document snapshot of size/400972
                        time:   [2.7905 ms 2.7935 ms 2.7969 ms]
                        change: [-0.4386% -0.1439% +0.2667%] (p = 0.46 > 0.05)
                        No change in performance detected.

Benchmarking [B4.1] Apply real-world editing dataset/259778: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 54.3s.
[B4.1] Apply real-world editing dataset/259778
                        time:   [5.4535 s 5.4993 s 5.5502 s]
                        change: [-2.0640% -0.8306% +0.3248%] (p = 0.20 > 0.05)
                        No change in performance detected.
```

tests-wasm/y-map.tests.js
chanced: @Horusiath could you not use the [Drop trait](https://doc.rust-lang.org/std/ops/trait.Drop.html)?

Waidhoferj: @Horusiath do iterators by reference block operations inside the loop that need a mutable reference to that transaction? That is, does the iterator object hold onto the reference until it is garbage collected?

Horusiath: @chanced The issue is not in the drop trait, but in a fact, that drop/free must be called manually in on an iterator created by yawsm to take an effect. Since iterator is created implicitly by the JS for loop, you never have a reference that you can call a `free()` on.

@Waidhoferj Yes, they do. Since iterator takes a readonly transaction reference, only readonly operations are allowed. Also since it never released, it means that you no longer be able to do anything mutating. I'd advice to check when pyo3 is calling Drop for you: does it have to be called explicitly or is it something automatic in specific cases.

Waidhoferj: PyO3 uses a `PyRef` class, which is similar to an `Rc<RefCell<T>>` in its implementation. It calls the `Drop` trait automatically, but I believe it is batched with Python's gc pass, so there aren't true sequential consistency guarantees.

Horusiath: What I did in ywasm eventually was to fallback from using iterators to return a full arrays/maps instead. They are created eagerly, so there's no need to keep on having a transaction for longer than it's necessary to build an array. This way all transaction references can be freed at the end of the method call.

dmonad (approved):
Looks good :+1:

PR #200, opened 2022-10-28
Fixed Update::integrate offset calculation

Horusiath:
Related to https://github.com/y-crdt/yrb/issues/45


yrs/src/update.rs
```diff
@@ -609,7 +609,7 @@ impl Update {
                                 } else {
                                     // if (currWrite.struct.id.clock + currWrite.struct.length >= curr.id.clock) {
                                     let diff =
-                                        curr_write_last as i32 - curr_block.id().clock as i32;
+                                        curr_write_last as i32 - curr_block.id().clock as i32 - 1;
```
Horusiath, line 612: How this variable is computed in Yjs:

```js
const diff = currWrite.struct.id.clock + currWrite.struct.length - curr.id.clock
```

Thing is that the difference between `struct.id.clock + struct.length` and `struct.last_id.clock` is always `1`.

eliias:
@Horusiath can confirm that the issue is fixed with 0.12.1 (y-rb 0.3.2 includes 0.12.1 and a test case).

PR #202, opened 2022-10-31
New type API

Horusiath:
Merge after #199

PR #210, opened 2022-11-16
Xml fragment as root level type

Horusiath:
Follow up to #202

- Introduce `XmlFragmentRef` as root level type.
- Remove `XmlTextRef` as a root level type - it should not be there anyway.
- Change `Doc::get_xml_element(name)` - now the `name` becomes a tag value of the root level `XmlElementRef`.
- Fix string generated by `XmlTextRef` - now it properly renders the formatting attributes as xml opening/closing tags with paramters.
- Renamed `to_string` &rarr; `get_string` - if a yrs trait handling this method was not imported, Rust compiler gave a wrong hint: it was mistaking yrs `to_string` method with core stdlib trait `ToString` which would lead to user confusion.


dmonad (approved):
Hurray, XML Fragments! :partying_face: (ping @stellit)

PR #226, opened 2022-12-13
Fix embeds V2 decoding

zaynetro:
Embeds are encoded to JSON but read as if there were binary encoded.

```
thread 'types::text::test::embed_with_attributes' panicked at 'Unable to read Any content'
```


Horusiath (approved):

Horusiath:
Thanks @zaynetro for fixing this issue. 👍

Issue #260, opened 2023-02-03
XMLFragment Treewalker leads to infinite loop


stellit:
Description:
with the following given base64url encoded binary, the tree walk ran into infinite loop:

[...]

PR #262, opened 2023-02-07
fixed infinite loop in deleted XML nodes

Horusiath:
This PR fixes bug reported in #260 .

### DIFF
--- a/yrs/src/types/xml.rs
+++ b/yrs/src/types/xml.rs
@@ -938,21 +938,29 @@ where
 
     /// Tree walker used depth-first search to move over the xml tree.
     fn next(&mut self) -> Option<Self::Item> {
+        fn try_descend(item: &Item) -> Option<&Block> {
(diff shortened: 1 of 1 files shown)


yrs/src/types/xml.rs
```diff
@@ -938,21 +938,29 @@ where
...
+
+            None
+        }
+
         let mut result = None;
         let mut n = self.current.take();
         if let Some(current) = n {
             if !self.first_call || current.is_deleted() {
                 while {
-                    if let ItemContent::Type(t) = &current.content {
```
Horusiath, line 946: Original source of an issue: we only alter node in while loop if its content type was `Type`. If it was i.e. `Deleted` we never updated the node nor break from the loop. The correct behavior is to descend on non-deleted `Type`, otherwise try to move right, then up to a parent hierarchy, even if current node was deleted.

PR #266, opened 2023-02-14
Slice does not always fit block range (adds move_range_to test from ypy)

stefanw:
While working on an upgrade of `ypy` to `yrs` v0.16.2 I noticed one of the `ypy` tests failing on a `debug_assert!` in `yrs`.

I recreated the [Python test case](https://github.com/y-crdt/ypy/blob/0df9bfe012b79d6e4f4d0c85e95a4bb59fd444c3/tests/test_y_array.py#L254) in Rust in `yrs` and it fails on this line:

https://github.com/y-crdt/y-crdt/blob/31b309868006d14c47c5deac6875979245de6409/yrs/src/moving.rs#L106

It fails while doing a `to_json` on an array [after the "Move 3-5 to 5" part](https://github.com/stefanw/y-crdt/blob/test-move_range_to/yrs/src/types/array.rs#L1479-L1482) of the test.

If you remove previous parts the tests often pass though! So it looks like this particular sequence of transactions causes the failure. Sorry that it's not a minimal test case but I really wouldn't know where to start. I hope it's helpful for tracking down the bug behind it.

The CI tests pass because they do `cargo test --release` and the `debug_assert!` is not triggered then.

### DIFF
--- a/yrs/src/types/array.rs
+++ b/yrs/src/types/array.rs
@@ -1349,6 +1349,228 @@ mod test {
     }
 
     #[test]
+    fn move_range_to() {
+        let doc = Doc::with_client_id(1);
+        let arr = doc.get_or_insert_array("array");
+        // Move 1-2 to 4
+        {
+            let mut txn = doc.transact_mut();
+            let arr_len = arr.len(&txn);
+            arr.remove_range(&mut txn, 0, arr_len);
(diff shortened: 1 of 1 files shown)


Horusiath:
Thanks for the test case. We've still looking for cases like that, which is why moving ranges is still considered experimental (moving single elements is fine though).

PR #269, opened 2023-02-27
Fix some inconsistencies in the Update::merge_updates method with Y.js.

### LINKED ISSUE #268: The implementation of merge_updates is inconsistent with Y.js.
1. lost update
2. duplicate updates

## lost update

```
==> u1:  {
    blocks: UpdateBlocks {
        clients: {
            1: [
                Item(<1#0>, len: 3, parent: test: 'aaa'),
                Item(<1#3>, len: 3, origin-r: <1#0>: 'aaa'),
            ],
        },
    },
}
==> u2:  {
    blocks: UpdateBlocks {
        clients: {
            2: [
                Item(<2#0>, len: 2, parent: test: 'bb'),
                Item(<2#2>, len: 1, origin-l: <2#1>: 'b'),
                Item(<2#3>, len: 3, origin-l: <2#1>, origin-r: <2#2>: 'bbb'),
            ],
        },
    },
}
```

After executing `let u12 = Update::merge_updates(vec![u1, u2]);`:

```
==> u12:  {
    blocks: UpdateBlocks {
        clients: {
            1: [
                Item(<1#0>, len: 3, parent: test: 'aaa'),
                Item(<1#3>, len: 3, origin-r: <1#0>: 'aaa'),
            ],
            2: [
                Item(<2#3>, len: 3, origin-l: <2#1>, origin-r: <2#2>: 'bbb'),
            ],
        },
    },
}
```

Two updates are lost: `Item(<2#0>, len: 2, parent: test: 'bb'),` and `Item(<2#2>, len: 1, origin-l: <2#1>: 'b'),`.

To reproduce the issue, modify the unit test method `update_merge` in the update.rs file as follows:

```rust
        txt1.insert(&mut t1, 0, "aaa");
        txt1.insert(&mut t1, 0, "aaa");

        // Change the following two lines of code
        // txt1.insert(&mut t1, 0, "bbb");
        // txt1.insert(&mut t1, 2, "bbb");
        // to:
        txt2.insert(&mut t2, 0, "bbb");
        txt2.insert(&mut t2, 2, "bbb");
```

## duplicate updates

When I merge the same update twice, I expect to get one `Item`, but currently I get two, one complete `Item` and one with truncated content.

```rust
==> update:  {
    blocks: UpdateBlocks {
        clients: {
            1: [
                Item(<1#0>, len: 3, parent: test: 'aaa'),
            ],
        },
    },
}
```

After executing `Update::merge_updates(vec![u1, u2])` (u1==u2):

```rust
==> merged_update:  {
    blocks: UpdateBlocks {
        clients: {
            1: [
                Item(<1#0>, len: 3, parent: test: 'aaa'),
                Item(<1#2>, len: 1, parent: test, origin-l: <1#1>, left: <1#0>: 'a'),
            ],
        },
    },
}
```

hamflx:
fix #268 

changes:

1. Modify the sorting of lazy_struct_decoders to ordering.reverse() to meet the requirement of descending order sorting by clientId, and resolve the issue of lost updates (unittest: modified `update_merge`).
2. Refactor the `if let .. {} else { continue/break }` structure to `let .. = match { _ => break/continue }` to reduce indentation and make the code more readable.
3. Replace `curr_write.take()` and `curr.take()` with references. The former moves ownership, causing subsequent calculations to lose access to the corresponding state.
4. To reference the block of the current iteration, I removed the `impl Iterator for Memo` and replaced it with the `next_item()` method, the former will not be available in the next iteration after the `next` method is called in the current iteration of `loop`, if the processing logic is not hit. And in the `move_next` method, the current iteration is saved to `current`.
5. Add a new parameter `ignore_skip` to IntoBlocks to ignore blocks of type skip when iterating over blocks. In the update generated by Y.js, a structure that is not received may be replaced by `Skip` and subsequently processed, which would result in missing parts of the content if `Skip` is not skipped here.
6. Replace "lazy_struct_decoders.pop_front()" with "lazy_struct_decoders.iter_mut().next()" so that multiple clients' data can be included in a single update (unittest: `test_multiple_clients_in_one_update`).
7. Replace `let diff == curr_write_last as i32 - curr_block.id().clock as i32 - 1;` with `let diff = curr_write_last.saturating_sub(curr_block.id ().clock);` to avoid -1 creating a redundant structure.

### DIFF
--- a/yrs/src/doc.rs
+++ b/yrs/src/doc.rs
@@ -1250,7 +1250,7 @@ mod test {
         let a = acc.clone();
         let _sub = d1.observe_update_v1(move |_: &TransactionMut, e| {
             let u = Update::decode_v1(&e.update).unwrap();
-            for mut block in u.blocks.into_blocks() {
+            for mut block in u.blocks.into_blocks(false) {
                 match block.as_block_ptr().as_deref() {
                     Some(Block::Item(item)) => {
                         if let ItemContent::String(s) = &item.content {
--- a/yrs/src/update.rs
+++ b/yrs/src/update.rs
@@ -1109,11 +1109,42 @@ mod test {
         txt.insert(&mut tr, 0, "aaa");
 
         let binary = tr.encode_update_v1();
-        let u1 = Update::decode(&mut DecoderV1::new(Cursor::new(binary.as_slice()))).unwrap();
-        let u2 = Update::decode(&mut DecoderV1::new(Cursor::new(binary.as_slice()))).unwrap();
-        let u3 = Update::decode(&mut DecoderV1::new(Cursor::new(binary.as_slice()))).unwrap();
+        let u1 = decode_update(&binary);
+        let u2 = decode_update(&binary);
+        let u3 = decode_update(&binary);
 
         let merged_update = Update::merge_updates(vec![u1, u2]);
         assert_eq!(merged_update, u3);
+    }
+
+    #[test]
+    fn test_multiple_clients_in_one_update() {
+        let binary1 = {
+            let doc = Doc::with_client_id(1);
+            let txt = doc.get_or_insert_text("test");
+            let mut tr = doc.transact_mut();
+            txt.insert(&mut tr, 0, "aaa");
+            tr.encode_update_v1()
+        };
+        let binary2 = {
+            let doc = Doc::with_client_id(2);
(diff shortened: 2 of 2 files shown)


Horusiath:
Thanks for the contribution. I'll try to make a release with the fix on the next day.

Horusiath (approved):

Issue #186, opened 2022-09-08
YArray's iterator hides elements after calling move_to 


zaynetro:
1. Build an array
2. Move last element to be first
3. Access element directly --> order is correct
4. Iterate over array --> all elements after first are missing
5. Push to array --> see all elements again

```rust
fn read_chars(txn: &mut yrs::Transaction) -> Vec<String> {
    txn.get_array("chars")
        .iter()
        .map(|v| v.to_string())
        .collect::<Vec<_>>()
}

let doc = yrs::Doc::new();
let txn = &mut doc.transact();

// Create an array
let ychars = txn.get_array("chars");
ychars.push_back(txn, "A");
ychars.push_back(txn, "B");

assert_eq!(vec!["A".to_string(), "B".to_string()], read_chars(txn));

// Move last item to be the first
ychars.move_to(txn, 1, 0);

// Accessing element directly works
assert_eq!(
    Some("A".to_string()),
    txn.get_array("chars").get(1).map(|v| v.to_string())
);

// But this fails with:
// 'test_yarray_move' panicked at 'assertion failed: `(left == right)`
// left: `["B", "A"]`,
// right: `["B"]`
assert_eq!(vec!["B".to_string(), "A".to_string()], read_chars(txn));

// But after we push to the same yarray then things recover.
ychars.push_back(txn, "C");
assert_eq!(
    vec!["B".to_string(), "A".to_string(), "C".to_string()],
    read_chars(txn)
);
```

Horusiath:
Thanks for reporting the bug. I think, we already noticed it some time ago. I'll take a look on how to fix this.

appflowy:
Hi @Horusiath . I have the same issue in `0.16.4`.

appflowy:
Here is my test demo:
```rust
#[test]
    fn test() {
        let doc = Doc::with_client_id(1);
        let array = doc.get_or_insert_array("array");
        let mut txn = array.transact_mut();
        array.insert(&mut txn, 0, "1");
        array.insert(&mut txn, 1, "2");
        array.insert(&mut txn, 2, "3");
        assert_eq!(array.iter(&txn).count(), 3);
        drop(txn);

        let txn = array.transact();
        let str_array = array
            .iter(&txn)
            .map(|v| v.to_string(&txn))
            .collect::<Vec<String>>();
        assert_eq!(str_array, vec!["1", "2", "3"]);
        drop(txn);

        let mut txn = array.transact_mut();
[...]

PR #282, opened 2023-03-27
Fixed - BlockIter ends prematurely if last array element has been moved

Horusiath:
Related #186

### DIFF
--- a/yrs/src/block_iter.rs
+++ b/yrs/src/block_iter.rs
@@ -44,7 +44,7 @@ impl BlockIter {
 
     #[inline]
     pub fn finished(&self) -> bool {
-        self.reached_end || self.index == self.branch.content_len
+        (self.reached_end && self.curr_move.is_none()) || self.index == self.branch.content_len
     }
 
     #[inline]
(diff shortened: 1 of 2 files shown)


yrs/src/block_iter.rs
```diff
@@ -371,49 +371,67 @@ impl BlockIter {
...
+                }
+            } else if self.curr_move.is_some() {
+                // reached end but move stack still has some items,
+                // so we try to pop move frames and move on the
+                // first non-null right neighbor of the popped move block
+                while let Some(Block::Item(mov)) = self.curr_move.as_deref() {
+                    next_item = mov.right;
+                    self.pop(txn);
+                    if next_item.is_some() {
+                        self.reached_end = false;
```
Horusiath, line 427: This is the clue. If we moved last block in an array, simply trying to move to `block.right` resulted in reaching no neighbour and setting `reached_end = true`. But if move stack is not empty, it means that we may just need to pop the moved stack frames and try right neighbour from there. We're popping recursively until first non-null right neighbour is found. 

/cc @dmonad

PR #286, opened 2023-03-29
add Result for Encode trait

darkskygit:
this pr aims to add `Result<(), lib0::error::Error>` return value to the `encode` function of `Encode` trait

this allows developers to safely handle errors when encountering erroneous data

after this pr is merged, I plan to modify the panic in yrs to a `YrsError` structure generated by thiserror in the next pr, which can make the error message clearer

### DIFF
--- a/yffi/src/lib.rs
+++ b/yffi/src/lib.rs
@@ -850,7 +850,10 @@ pub unsafe extern "C" fn ytransaction_state_vector_v1(
 
     let txn = txn.as_ref().unwrap();
     let state_vector = txn.state_vector();
-    let binary = state_vector.encode_v1().into_boxed_slice();
+    let binary = state_vector
+        .encode_v1()
+        .expect("failed to encode state vector using lib0 v1 encoding")
+        .into_boxed_slice();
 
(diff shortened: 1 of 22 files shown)


Horusiath:
@darkskygit thanks for your contribution. I'll check it in more detail, once I return from the vacation.

While I was in favour of using results and `std::io::Write`-based encoding once I joined a project, we eventually decided to no go this way. It wasn't an accidental decision. It may be a good time to reevaluate it and make changes, however please keep in mind that doing this kind of PR without prior proposal may turn into a wasted effort from your side.

darkskygit:
> @darkskygit thanks for your contribution. I'll check it in more detail, once I return from the vacation.
> 
> While I was in favour of using results and `std::io::Write`-based encoding once I joined a project, we eventually decided to no go this way. It wasn't an accidental decision. It may be a good time to reevaluate it and make changes, however please keep in mind that doing this kind of PR without prior proposal may turn into a wasted effort from your side.

Thank you for your reply, if you have any suggestions for revision, please feel free to let me know, I will make revisions as soon as possible.

Issue #295, opened 2023-05-31
Panic when calling encode_update_v2 on TransactionMut


stellit:
Hi!

There is a mysterious behavior found with the TransactionMut.encode_update_v2() method.

In the given test code, if line `[A]` with the first `encode_update_v2` commented out, the code will crashes at the line `[B]`. But if enable the call `[A]`, `[B]` will pass smoothly. I've also attached the same logic that tested in yjs:

Here is the code in rust:
```rust
#[test]
fn test_yrs_issue() {
    let before = "AAEAC4HX78kQxu2Y3AgEBAAEAQEACygDJwAEAAcAKAMIowGOAW5vdGUuZ3VpZG5vdGVHdWlkbm90ZS5vd25lcm93bmVybm90ZS50eXBlbm90ZVR5cGVub3RlLmNyZWF0ZVRpbWVjcmVhdGVUaW1lbm90ZS50aXRsZXRpdGxlMXByb3NlbWlycm9ybGlua25vdGVndWlkc2xvdEd1aWR0eXBlbGlua1R5cGVjaGlsZHJlbmIJCAoFCQgPSgAFAQsIBAgESAABCQEEAAABAAADAQICAwJBCAIEAHcWYWNGeFVZRExSaGVya0phQmVjc2MzZ32IOX0Adw0xNjg1NTMwMjg2NjU2CQB3FkJDZFFwcHdFVDBpUlZCUS04RVcyV2d3FmpyRW1JTXBWVGUtY3JOMlZMM2NhSFF3CGxpbmtub3RldwExdgIEZ3VpZHcWQkNkUXBwd0VUMGlSVkJRLThFVzJXZwhzbG90R3VpZHcWanJFbUlNcFZUZS1jck4yVkwzY2FIUXcWQkNkUXBwd0VUMGlSVkJRLThFVzJXZwA";
    let update = "AAAAAAAAAQAAAAAAAca2jK4EAQIG";
    let raw_doc = decode_base64_url_safe(before).unwrap();
    let update = decode_base64_url_safe(update).unwrap();
    let doc = empty_doc(true);
    {
        let mut txn = doc.transact_mut();
        txn.apply_update(Update::decode_v2(&raw_doc).unwrap());
        // txn.encode_update_v2(); // [A] first encode_update_v2() call
    }
    {
        let b = doc.get_or_insert_array("b");
        let txn = doc.transact();
        println!("before update: {}", b.to_json(&txn));
    }
    {
[...]

PR #296, opened 2023-06-09
fixed DeleteSet encoding over unsquashed DeleteSet

Horusiath:
Related to #295

This PR is a brute force fix for issue with V2 encoding, when `DeleteSet` is being serialized before being squashed.

### DIFF
--- a/yrs/src/id_set.rs
+++ b/yrs/src/id_set.rs
@@ -162,6 +162,29 @@ impl IdRange {
         }
     }
 
+    fn is_squashed(&self) -> bool {
+        match self {
+            IdRange::Continuous(_) => true,
+            IdRange::Fragmented(ranges) => {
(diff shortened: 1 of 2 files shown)


yrs/src/id_set.rs
```diff
@@ -216,17 +254,12 @@ impl Default for IdRange {
...
-            IdRange::Fragmented(ranges) => {
-                encoder.write_var(ranges.len() as u32);
-                for range in ranges.iter() {
-                    range.encode(encoder);
-                }
-            }
+        if self.is_squashed() {
+            self.encode_raw(encoder)
+        } else {
+            let mut clone = self.clone();
```
Horusiath, line 260: The fix is basically - check if `DeleteSet` was squashed and if not, clone it and squash before encoding.

PR #307, opened 2023-07-19
Expose a `root_keys` iterator on a transaction

paulgb:
As far as I can tell, there is currently no way to iterate over keys in a document without calling `BranchPtr::into<Value>`. This is unfortunate since `BranchPtr::into` [will panic](https://github.com/y-crdt/y-crdt/blob/50d4cb6372d73e66d9243a38efe1668fda486b2b/yrs/src/types/mod.rs#L276) if that key is not already associated with a type.

This PR exposes an iterator that can be used to read the keys without doing anything with the associated values. The keys can then be used with `Transcation::get_array`, `Transaction::get_map`, etc. to access views of that data. This is useful for reflection when the set of keys is not known in advance and needs to be determined programatically.

### DIFF
--- a/yrs/src/transaction.rs
+++ b/yrs/src/transaction.rs
@@ -86,6 +86,12 @@ pub trait ReadTxn: Sized {
         RootRefs(store.types.iter())
     }
 
+    /// Returns an iterator over top level (root) keys available in current [Doc].
+    fn root_keys(&self) -> RootKeys {
+        let store = self.store();
+        RootKeys(store.types.keys())
+    }
+
     /// Returns a collection of globally unique identifiers of sub documents linked within
     /// the structures of this document store.
     fn subdoc_guids(&self) -> SubdocGuids {
@@ -902,6 +908,19 @@ impl<'doc> Iterator for RootRefs<'doc> {
         Some((key, ptr.into()))
     }
 }
+
+pub struct RootKeys<'doc>(std::collections::hash_map::Keys<'doc, Arc<str>, Box<Branch>>);
+
+impl<'doc> Iterator for RootKeys<'doc> {
+    type Item = &'doc str;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let key = self.0.next()?;
+        let key = key.as_ref();
+        Some(key)
+    }
+}
+
 
 #[derive(Default)]
 pub struct Subdocs {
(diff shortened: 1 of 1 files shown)


Horusiath (changes requested):
I think this is exposing too much - untyped branches should not be exposed lightly. We could just return `fn root_keys(&self) -> impl Iterator<Item=Arc<str>>` and let the users access the keys they need themselves.

paulgb:
Unfortunately returning an `impl` is not possible here because it’s a trait. `std::collections::hash_map::Keys` does not actually expose the `Box<Branch>`, but requires the type information stick around.

I've modified the code to wrap the iterator in an opaque `RootKeys` type to hide the type information from the caller, is that preferable? Another approach is to return `&str` rather than clone the `Arc`, shown in b2dca1c465a7ff3fc29743ee9007dd1ac9403032.

Horusiath (approved):

Horusiath:
Closing this one, as we already implemented this feature a while ago.

paulgb:
Thanks for the update, @Horusiath. Can you point me to the proper way to iterate the root keys (names) regardless of whether they are typed? I have tried on 0.17.1, and `root_refs` will still panic if a root type is not known for the current document. Here's what I tried.

```rust
use yrs::updates::decoder::Decode;
use yrs::{Doc, ReadTxn, StateVector, Text, Transact, Update};

fn main() {
    let doc = Doc::new();
    let text = doc.get_or_insert_text("foo");

    {
        text.insert(&mut txn, 0, "abc");
    }

    let update = doc.transact().encode_state_as_update_v1(&StateVector::default());
    let doc2 = Doc::new();
    
    {
        let mut txn = doc2.transact_mut();
        let update = Update::decode_v1(&update).unwrap();
        txn.apply_update(update);
    }

    {
        let txn = doc2.transact();
        for root_ref in txn.root_refs() {
            println!("root_ref: {:?}", root_ref);
        }
    }
}

```

Horusiath:
Hi Paul. This behaviour will be changed from panics to errors - as it touches other parts of the system. However in general it's an anti pattern to create a document without initialising its root types - updates encoding format doesn't propagate type of a root refs.

paulgb:
Thanks Bartosz, it looks like #357 should solve the problem for me. I understand that it is usually an anti-pattern, but it is necessary for debugging/introspection tools that don't require the user to specify a schema in advance.

PR #313, opened 2023-08-25
Reduced ClientID to 63bit/31bit

Horusiath:
This is an experiment where we're instructing Rust compiler to make better use of `ClientID`. Atm. it's just alias for `u64`. The thing is that when used as is it causes an explosion of memory usage. Consider following calculations:

```rust
type ClientID = u64; // mem. used 8B

// ID - memory used: 16B (8B for client_id, 4B for clock and 4B padding)
struct ID { client_id: ClientID, clock: u32 }
```

For `Option<ID>` each such structure uses **24B** (Option case adds 8B). These are all assuming x64 architecture. Now each Item uses two `Option<ID>` for it's neighbors, and `ID` for itself: 24 + 24 + 16 = **64B** just for these 3 fields.

One issue is that Option adds 8 bytes of extra overhead, even though 1bit should be sufficient - Rust does that since we're talking about value types. However Rust also has concept of `NonZeroU32`/`NonZeroU64` types, which are basically `u32`/`u64` that are guaranteed to never be equal to 0. Rust compiler knowing that can specialize `0` to represent `None` in any struct that defines field with such type: our `ID` included. This reduces the size of `Option<ID>` by 8 bytes (or at least 16 bytes per `Item`).

Furthermore if we use 31bit client IDs, we can also remove padding added by the compiler, which means that for `ClientID = NonZeroU32`, corresponding `Option<ID>` size drops from **24B** &rarr; **8B** (1/3 of its current size).

# Numbers

What does that mean in overall perspective? I'm running our trace editing tests (they are traces of real-life users editing their documents, taken from Sephs Gentle repository) with tracing allocator turned on. Here are the numbers:

Current main (our baseline):

|                 | elapsed (ms) | mem used (bytes) |
|-----------------|---------|------------------|
| automerge-paper | 16016 | 2446298 |
| friendsforever_flat | 56 | 672210 |
| seph-blog1 | 2025 | 2187595 |
| sveltecomponent| 9611 | 3136457 |
| rustcode | 339 | 991209 |

`ClientID` equal to `NonZeroU64` with highest bit set to make sure that ClientID zero will still work. The CPU time dropped, but at the same time overall memory used also dropped by 11%.

|                 | elapsed (ms) | mem used (bytes) |
|-----------------|---------|------------------|
| automerge-paper | 18079 | 2182994 |
| friendsforever_flat | 65 | 597402 |
| seph-blog1 | 2698 | 1946539 |
| sveltecomponent| 10681 | 2774249 |
| rustcode | 378 | 880161 |

`ClientID` equal to `NonZeroU32` (31 bit numbers) with highest bit set to make sure that ClientID zero will still work. The CPU time has not changed - probably since we add time on applying bit masks, but we get it in return since passing IDs by value can now be done in a single register mov - but at the same time overall memory used also dropped by 25-26%!

|                 | elapsed (ms) | mem used (bytes) |
|-----------------|---------|------------------|
| automerge-paper | 16065| 1831914 |
| friendsforever_flat | 56 | 497650 |
| seph-blog1 | 2046| 1625123 |
| sveltecomponent| 9615 | 2291297 |
| rustcode | 332| 732089 |

### DIFF
--- a/yffi/src/lib.rs
+++ b/yffi/src/lib.rs
@@ -298,7 +298,7 @@ impl Into<Options> for YOptions {
             Some(str)
         };
         Options {
-            client_id: self.id as ClientID,
+            client_id: ClientID::new(self.id),
             guid,
             collection_id,
             skip_gc: if self.skip_gc == 0 { false } else { true },
@@ -312,7 +312,7 @@ impl Into<Options> for YOptions {
 impl From<Options> for YOptions {
     fn from(o: Options) -> Self {
         YOptions {
-            id: o.client_id,
+            id: o.client_id.into(),
             guid: CString::new(o.guid.as_ref()).unwrap().into_raw(),
             collection_id: if let Some(collection_id) = o.collection_id {
                 CString::new(collection_id).unwrap().into_raw()
@@ -412,7 +412,7 @@ pub extern "C" fn ydoc_new_with_options(options: YOptions) -> *mut Doc {
 #[no_mangle]
 pub unsafe extern "C" fn ydoc_id(doc: *mut Doc) -> u64 {
     let doc = doc.as_ref().unwrap();
-    doc.client_id()
+    doc.client_id().into()
 }
 
 /// Returns a unique document identifier of this [Doc] instance.
@@ -3549,7 +3549,7 @@ impl YStateVector {
         let mut client_ids = Vec::with_capacity(sv.len());
         let mut clocks = Vec::with_capacity(sv.len());
         for (&client, &clock) in sv.iter() {
-            client_ids.push(client as u64);
+            client_ids.push(u64::from(client));
             clocks.push(clock as u32);
         }
 
--- a/yrs/src/block.rs
+++ b/yrs/src/block.rs
@@ -9,11 +9,16 @@ use crate::updates::encoder::{Encode, Encoder};
 use crate::utils::OptionExt;
 use crate::*;
 use lib0::any::Any;
+use lib0::decoding::Read;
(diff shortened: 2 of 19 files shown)


dmonad:
Nice!

> ClientID equal to NonZeroU64 with highest bit set to make sure that ClientID zero will still work. The CPU time dropped, but at the same time overall memory used also dropped by 11%.

It is interesting to note that this approach will double the probability of collision, which for `NonZeroU32` is already quite high if there are a few hundred users.

I think for this reason, we shouldn't investigate U32 clientids further. The space is simply too limited.

I think we can limit clientids to exactly 53bit, which can be represented as a float as well. Maybe the remaining bits can be used for future optimizations.

dmonad:
Just a technical note, although I still don't like U32:

ClientId 0 is quite useful. Banning it would be counter-intuitive. But maybe the highest integer is not that important. Hence, you could store numbers as the inverse (`~0` = `2^32-1`), thus banning the highest integer. This doesn't have a significant impact on collision probability. But we'd have to disallow that number everywhere then.

Horusiath:
@dmonad it's not banning client id `0` - what I'm doing is internally every client gets highest bit (32) set, then we clear that bit when serialisation needs to happen. This way we can use client ID = 0 (and even exchange it with Yjs), while still ensuring Rust compiler that 0 is free to be used. Off it also means that the supported client ID is effectively 31 bit.

dmonad:
Sure, I understand that. I was entertaining the idea to ban the highest integer instead of limiting ourselves to u31 which doubles collision probability.

PR #315, opened 2023-08-31
lib0 binary encoding: encode small ints as integer not big int

### LINKED ISSUE #314: Int always became BigInt.
```main.rs
use ::lib0::serde::to_any;
use ::serde_json::{Result, Value};

fn main() {
    bigint_demo().unwrap();
}

fn bigint_demo() -> Result<()> {
    let data = r#"
        {
            "App": {
              "device": {
                "size": {
                  "width": 1440,
                  "height": 1024
                }
              }
            }
        }"#;

    // Parse the string of data into serde_json::Value.
    let v: Value = serde_json::from_str(data)?;

    // Access parts of the data by indexing with square brackets.
    println!("v is {:?}", v);

    println!("v to_any {:?}", to_any(&v).unwrap());

    Ok(())
}
```

which output:
image

when the client-side yjs, we will have `1440n` which is weird.

Horusiath:
Fixes #314

Basically we did JS safe int bounds check for `Any::Number` but not for `Any::BigInt`.

### DIFF
--- a/lib0/src/any.rs
+++ b/lib0/src/any.rs
@@ -36,6 +36,7 @@ impl std::fmt::Display for Number {
 macro_rules! impl_float {
     ($t:ty) => {
         impl From<$t> for Number {
+            #[inline]
             fn from(v: $t) -> Number {
                 Number::Float(v as f64)
             }
@@ -46,6 +47,7 @@ macro_rules! impl_float {
 macro_rules! impl_number {
     ($t:ty) => {
         impl From<$t> for Number {
+            #[inline]
             fn from(v: $t) -> Number {
                 Number::Int(v as i64)
             }
@@ -120,9 +122,9 @@ pub enum Any {
     Undefined,
     Bool(bool),
     Number(Number),
-    String(String),
-    Buffer(Vec<u8>),
-    Array(Vec<Any>),
+    String(Box<str>),
+    Buffer(Box<[u8]>),
+    Array(Box<[Any]>),
     Map(Box<HashMap<String, Any>>),
 }
 
@@ -146,10 +148,7 @@ impl Any {
             // CASE 120: boolean (true)
             120 => Any::Bool(true),
             // CASE 119: string
-            119 => {
-                let str = decoder.read_string()?;
-                Any::String(str.to_string())
-            }
+            119 => Any::String(Box::from(decoder.read_string()?)),
             // CASE 118: Map<string,Any>
             118 => {
                 let len: usize = decoder.read_var()?;
@@ -167,10 +166,10 @@ impl Any {
                 for _ in 0..len {
                     arr.push(Any::decode(decoder)?);
                 }
-                Any::Array(arr)
+                Any::Array(arr.into_boxed_slice())
             }
             // CASE 116: buffer
-            116 => Any::Buffer(decoder.read_buf()?.to_owned()),
+            116 => Any::Buffer(Box::from(decoder.read_buf()?)),
             _ => return Err(Error::UnexpectedValue),
         })
     }
@@ -407,6 +406,7 @@ impl std::fmt::Display for Any {
 }
 
 impl From<bool> for Any {
+    #[inline]
     fn from(v: bool) -> Any {
         Any::Bool(v)
     }
@@ -414,7 +414,7 @@ impl From<bool> for Any {
 
 impl From<Vec<u8>> for Any {
     fn from(v: Vec<u8>) -> Any {
-        Any::Buffer(v)
+        Any::Buffer(v.into_boxed_slice())
     }
 }
 
@@ -427,7 +427,7 @@ where
         for value in v {
             array.push(value.into())
         }
-        Any::Array(array)
+        Any::Array(array.into_boxed_slice())
     }
 }
 
@@ -448,20 +448,23 @@ impl<T> From<T> for Any
 where
     T: Into<Number>,
 {
+    #[inline]
     fn from(value: T) -> Self {
         Any::Number(value.into())
     }
 }
 
 impl From<String> for Any {
+    #[inline]
     fn from(v: String) -> Any {
         Any::String(v.into())
     }
 }
 
 impl From<&str> for Any {
+    #[inline]
     fn from(v: &str) -> Any {
-        Any::String(v.to_string())
+        Any::String(Box::from(v))
     }
 }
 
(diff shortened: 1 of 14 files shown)


dmonad (commented):
This is not the behavior of lib0. What was once stored as BigInt should also be returned as a BigInt.

IndexXuan:
@Horusiath @dmonad May be we should improve the `to_any` method ?
https://github.com/y-crdt/y-crdt/blob/814c778a638ed6c323165320de2a8b894a2b051c/lib0/src/any.rs#L36

Horusiath:
@dmonad the thing is that JS doesn't recognize between different number types. Integers are `Number`, so are floating point numbers. In Rust `f32` and `f64` are stored as `Number(f64)` while any integer is stored as `BigInt(i64)`. 

IMO the problem is that we're sticking to JS type system, which is the one of the most convoluted and error-prone type systems out there. Our `Any` type should accept either `Any::Float(f64)` or `Any::Integer(i64)` - the problem is that without any adjustments this would break lib0 encoding.

I can think of 2 solutions:

1. The one proposed here (preferable).
2. Change conversion of `Into<Any>` for integer types to use `MAX_SAFE_INT` guard there ie. `let x: Any = 123.into()` would produce `Any::Number` while `let x: Any = i64::MAX.into()` would produce `Any::BigInt`. But this also mean that we loose the information if number passed from Rust was integer or floating point.

dmonad:
> @dmonad the thing is that JS doesn't recognize between different number types. Integers are Number, so are floating point numbers. In Rust f32 and f64 are stored as Number(f64) while any integer is stored as BigInt(i64).

I feel that Rusts automatic conversion feature leads to confusion. f64 can be stored as both `Any::Number` or `Any::BigInt`, depending on whether it is an integer or a floating point number. So maybe automatic conversion should convert `f64` always to `Any::Number` (which looks like the correct target) and always return `Any::Number(f64)`

The same goes for i64. 

@IndexXuan I don't understand what you are saying. If you are saying that Yjs represents numbers differently depending on whether you use it in the browser or node, please open a ticket in the Yjs repository. Please also give a full reproducible example.

Horusiath:
> The Any encoding format is a superset of JSON. JSON only supports encoding of floats. 

@dmonad JSON just uses numbers, without precising if they are float or not. Literally every single language that has that distinction, also distinguishes between different versions of numbers without implicit conversions.

I think, the most reasonable way would be to provide 3 number variants: `Float(f64)`, `Int(i64)` and `Bigint(i64)` . Float is always converted with tag `124` (float 32) or `123` (float 64), Int => always tag `125` (integer) and BigInt => always tag `122` (big int). Currently it's much more convoluted and even not symetric: https://github.com/y-crdt/y-crdt/blob/814c778a638ed6c323165320de2a8b894a2b051c/lib0/src/any.rs#L123-L146

Horusiath:
So I've adapted the PR with the following ideas:

1. Introduced `Number` enum with 3 cases: 
    - `Float(f64)` (for floating point numbers): encoded using either tag 123 (f64) or 124 (f32) depending on the precision check.
    - `Int(i64)` (for 53bit integers): always encoded using tag 125 (integer) and var int
    - `BigInt(i64)` (for biger integers): always encoded tag 122 (big int) and fixed i64
2. Replaced `Into<Any>` with `From<T> for Any`/`From<T> for Number` - in general Rust compiler auto-generates `Into<Any>` trait for each type that implements `From<T> for Any`, but not the other way around. It's more permissive this way. 
3. Also `From<T> for Number` implementations for `i64`/`isize` now can return either `Number::Int` or `Number::BigInt` depending on acutal value fitting the `MAX_SAFE_INTEGER` boundaries.
4. Replaced `Any::Buffer(Box<[u8]>)`&rarr;`Any::Buffer(Vec<u8>)`, `Any::String(Box<str>)`&rarr;`Any::String(String)` and `Any::Array(Box<[Any]>)`&rarr;`Any::Array(Vec<Any>)` as they were in the past - using raw boxes for collections is almost never the case in Rust.

dmonad:
Just for clarification of what I'm striving for, the JSON RFC states:

>  Since software that implements IEEE 754 binary64 (double precision) numbers [IEEE754] is generally available and widely used, good interoperability can be achieved by implementations that expect no more precision or range than these provide, in the sense that implementations will approximate JSON numbers within the expected precision.

This is why I made BigInt a special case. No more special cases are necessary in my opinion. It could be argued that we should remove BigInt.

The concept of `Int(i64)` for 53bit integers seems weird to me. 1. Users have to know and understand that. They have to make sure that they don't exceed the 53bit space. 2. All 53 bit integers fit into `Number(f64)` as well. So what is the benefit in separating them? 3. it is unclear how JavaScript and other "inferior languages" can separate these two concepts.

Horusiath:
@dmonad problem is that when you're working with typed languages you eventually need to work with integer or float values. This requires casting, that may be potentially invalid and resulting in a data loss. Casting `10.5 as i64` is valid, but the problem is that it will cause data loss without even informing user about it. You also can never be sure if such thing will happen, because ie. some update may produce correct result 4 out of 5 times, and then it turns out that one side introduced a breaking change in a data model that is silently ignored by other party (ie. changing `{amount: number}` from integer to fractional number that will be ok, until you actually start to use it for computation and produce different results on both ends).

dmonad:
That is why the JSON spec recommends limiting numbers to f64. I would assume that even serde automatically casts everything to f64.

How I see it, int53 is just a special case of float that we can encode more efficiently using variable length encoding. The result should still be f64. You can cast the result to your expected data type. Maybe that process could be more efficient, but we are not serving system engineers that understand everything about different number representations. We mainly serve people like me, simple-minded JavaScript developers that just want to store some "JSON".

Now you could argue that it would be more efficient to avoid casting an integer to float and back to integer. My arguments:

1. Performance is neglegible.
2. If you want to store a float that happens to be an integer (eg number 1.0), then I'd expect that the result is still a float and not an integer. I think the current implementation would automatically cast it to an integer. Confusing. 3. JavaScript doesn't even know the difference between float and int. That's why I'm pushing for compatibility so that everyone can talk about the same thing.

Horusiath:
I'm closing this PR. I've made changes in the #321, where I'm converting ints to `Any::Number(f64)` or `Any::BigInt(i64)` depending on if they reach beyond 53bits or not.

Issue #337, opened 2023-10-07
ywasm examples aren't working


rgbkrk:
The example in the readme has a few missing pieces namely that `beginTransaction` doesn't exist anymore and `transact` isn't being called below:

```typescript
import {YDoc, encodeStateVector, encodeStateAsUpdate, applyUpdate} from 'ywasm';

YDoc.prototype.transact = callback => {
    const txn = this.beginTransaction()
    try {
        return callback(txn)
    } finally {
        txn.free()
    }
}

const doc = new YDoc()
const text = doc.getText('name')

// append text to our collaborative document
text.insert(txn, 0, 'hello world')

// simulate update with remote peer
const remoteDoc = new YDoc()
const remoteText = remoteDoc.getText('name')

// in order to exchange data with other documents 
// we first need to create a state vector
const remoteSV = encodeStateVector(remoteDoc)
// now compute a differential update based on remote document's state vector
const update = encodeStateAsUpdate(doc, remoteSV)
// both update and state vector are serializable, we can pass them over the wire
// now apply update to a remote document
applyUpdate(remoteDoc, update)

const str = remoteText.toString(txn)
console.log(str)
```

I looked into an example inside of `ywasm/src/lib.rs` and it seems the interface is now:

```
// example.mjs
import {YDoc} from 'ywasm'

const doc = new YDoc()

const text = doc.getText('name')

const txn = doc.readTransaction()
try {
    text.insert(txn, 0, 'hello world')
} finally {
    txn.free()
}
```

However that errors out in both node and deno with a memory access out of bounds.

```
Stack trace:
RuntimeError: memory access out of bounds
    at dlmalloc::dlmalloc::Chunk::size::h48e91a2c5adb13a6 (wasm://wasm/00378db6:1:751711)
    at __rdl_realloc (wasm://wasm/00378db6:1:382950)
    at __rust_realloc (wasm://wasm/00378db6:1:750065)
    at __wbindgen_realloc (wasm://wasm/00378db6:1:745167)
    at passStringToWasm0 (file:///Users/kai/Library/Caches/deno/npm/registry.npmjs.org/ywasm/0.16.10/ywasm.js:119:15)
    at YText.insert (file:///Users/kai/Library/Caches/deno/npm/registry.npmjs.org/ywasm/0.16.10/ywasm.js:2109:26)
    at <anonymous>:7:8
```

This is all with `ywasm@0.16.10` from npm

Horusiath:
[...]

PR #353, opened 2023-11-17
Exposed transactions in ywasm observer callbacks

Horusiath:
This PR exposes transaction in observer callback in the ywasm API. 

Previously this was not possible because of the duality of Transaction behaviour: when created via `doc.readTransaction()` or `doc.writeTransaction()` it was a value, boxed and returned to JS API, while observer callbacks were always referring to already referenced transaction (not a value).

Unfortunately it's sometimes necessary to have transaction references in a callback ie. to check the origin of an update. This PR enables to expose both Transaction by ref and by value to JS API.

It also fixes outdated documentation issue mentioned in #337


davidbrochart:
Is it also how the transaction in a callback should be passed e.g. to ypy or pycrdt? I remember you talked about a general way that all language bindings could use.

Horusiath:
@davidbrochart My solution is basically this: https://github.com/y-crdt/y-crdt/blob/be70e50734c9c16d965ffba2ea1b9f2e0100c7d4/ywasm/src/lib.rs#L697-L739 (it's an enum that can hold either transaction by value or by reference).

It's similar to Rust's standard library `Cow` type but it has less constraints. I think there might be also other ways to achieve the same goal, but tbh. this seems to be right and working approach.

I've figured out, that this approach is very generic, but it's I couldn't justify its purpose to include it straight into yrs. It's just a 20+ lines of very simple code, so I think it's ok to just copy it around if necessary.

PR #363, opened 2023-12-04
fix: serialize i64

appflowy:
When an integer (e.g., `1` as `i64`) is converted to the `Any` type using `Any::from`, it is transformed into a `number` representation within the `Any` type.  For example, if you serialize `1` (as `i64`), you might expect to deserialize it back to an integer `1`, but instead, you get a floating-point `1.0`.

### DIFF
--- a/yrs/src/encoding/serde/ser.rs
+++ b/yrs/src/encoding/serde/ser.rs
@@ -715,4 +715,11 @@ mod test {
 
         assert_eq!(any, to_any(&any.clone()).unwrap())
     }
+
+    #[test]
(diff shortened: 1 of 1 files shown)


Horusiath:
Unfortunately we're bound for Yjs lib0 encoding: and therefore JavaScript type system, which doesn't recognise between int and float types. It means that if you have ie. `1` in Yjs, encode it as an update, pass to Yrs, then serialise and send back, JavaScript would expect `1` back. While int/float implicit conversions are supported by JavaScript, BigInt is encoded as a separate type, that doesn't implicitly convert to a number.

PR #369, opened 2023-12-18
chore: replace from with_capacity with try_reserve

appflowy:
Replace the use of `with_capacity` with `try_reserve` in collections specifically when decoding the `Update` to better handle potential memory allocation errors. For now, this change should be applied only in the context of `Update` decoding, without altering other instances of `with_capacity` usage. 

If needed, please let me know, and I can handle the remaining tasks

### DIFF
--- a/yffi/src/lib.rs
+++ b/yffi/src/lib.rs
@@ -1175,12 +1175,16 @@ pub const ERR_CODE_INVALID_JSON: u8 = 5;
 /// Error code: other error type than the one specified.
 pub const ERR_CODE_OTHER: u8 = 6;
 
+/// Error code: allocated buffer is too small to hold a decoded value.
+pub const ERR_NOT_ENOUGH_BYTES: u8 = 7;
+
 fn err_code(e: Error) -> u8 {
     match e {
         Error::VarIntSizeExceeded(_) => ERR_CODE_VAR_INT,
         Error::EndOfBuffer(_) => ERR_CODE_EOS,
         Error::UnexpectedValue => ERR_CODE_UNEXPECTED_VALUE,
         Error::InvalidJSON(_) => ERR_CODE_INVALID_JSON,
+        Error::NotEnoughBytes(_) => ERR_NOT_ENOUGH_BYTES,
     }
 }
 
--- a/yffi/src/lib.rs
+++ b/yffi/src/lib.rs
@@ -1175,8 +1175,8 @@ pub const ERR_CODE_INVALID_JSON: u8 = 5;
 /// Error code: other error type than the one specified.
 pub const ERR_CODE_OTHER: u8 = 6;
 
-/// Error code: allocated buffer is too small to hold a decoded value.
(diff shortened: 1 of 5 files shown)


Horusiath:
This is a little bit tricky. A fact that you've allocated enough space for blocks alone doesn't mitigate all the issues. You can have enough memory to allocate blocks, but still not enough to allocate block-related data.

appflowy:
Yes, refactoring is necessary for managing block-related data more effectively. In the current setup, decoding invalid data leads to a memory allocation error, which results in the immediate termination of the program. One approach I'm considering to prevent this issue is implementing 'Input Validation.' This would involve checking the data before decoding to ensure it adheres to the expected format and constraints, thereby avoiding the decoding of manifestly invalid data. Do you think this would be an effective solution, or do you have any other recommendations?

yffi/src/lib.rs
```diff
@@ -1175,12 +1175,16 @@ pub const ERR_CODE_INVALID_JSON: u8 = 5;
 /// Error code: other error type than the one specified.
 pub const ERR_CODE_OTHER: u8 = 6;
 
+/// Error code: allocated buffer is too small to hold a decoded value.
+pub const ERR_NOT_ENOUGH_BYTES: u8 = 7;
```
Horusiath, line 1179: Maybe `ERR_NOT_ENOUGH_MEMORY`?

yrs/src/encoding/read.rs
```diff
@@ -12,6 +13,9 @@ pub enum Error {
     #[error("while reading, an unexpected value was found")]
     UnexpectedValue,
 
+    #[error("failed to allocate memory: {0}")]
+    NotEnoughBytes(#[from] TryReserveError),
```
Horusiath, line 17: Same here, `NotEnoughMemory` seems to be better description of the problem.

Horusiath (changes requested):
If this is an approach to fight with malicious or corrupted updates, then we'd need to specify what is a max allowed update size and verify it after decoding the clients/block length.

But since it's hard to determine these limits before reading the entire update, the proactive approach with `try_reserve` seems to be plausible. My only note is an error name - please change it to better indicate that it's a memory allocation problem.

appflowy:
> If this is an approach to fight with malicious or corrupted updates, then we'd need to specify what is a max allowed update size and verify it after decoding the clients/block length.
> 
> But since it's hard to determine these limits before reading the entire update, the proactive approach with `try_reserve` seems to be plausible. My only note is an error name - please change it to better indicate that it's a memory allocation problem.

Got it. Will do later

appflowy:
@Horusiath Is there anything else I need to do for this pull request before it can be merged?

Horusiath:
@appflowy I see there are merge conflicts. I can take it over from here and merge it this week.

appflowy:
> @appflowy I see there are merge conflicts. I can take it over from here and merge it this week.

Done

Horusiath (approved):

Issue #386, opened 2024-02-09
Crash while applying a update that contains block of size 0


xyzhangfred:
Hi, we encountered a crash while using the C FFI of yrs. It was due to a block having the size of 0, (not sure how exactly), which leads to overflow when subtracted by 1 (at block_store.rs:452:22). We were able to 100% reproduce it with the same data, but unfortunately we cannot share the date here since it contains sensitive info. 
Here is the stack trace:
`
thread '<unnamed>' panicked at 'attempt to subtract with overflow', yrs/src/block_store.rs:452:22
stack backtrace:
   0: rust_begin_unwind
             at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/std/src/panicking.rs:593:5
   1: core::panicking::panic_fmt
             at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/core/src/panicking.rs:67:14
   2: core::panicking::panic
             at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/core/src/panicking.rs:117:5
   3: yrs::block_store::BlockStore::get_item_clean_end
             at /Users/fred_zoom/SourceCode/y-crdt/yrs/src/block_store.rs:452:22
   4: yrs::block::Item::repair
             at /Users/fred_zoom/SourceCode/y-crdt/yrs/src/block.rs:1448:25
   5: yrs::update::Update::integrate
             at /Users/fred_zoom/SourceCode/y-crdt/yrs/src/update.rs:220:33
   6: yrs::transaction::TransactionMut::apply_update
             at /Users/fred_zoom/SourceCode/y-crdt/yrs/src/transaction.rs:559:41
   7: ytransaction_apply
             at /Users/fred_zoom/SourceCode/y-crdt/yffi/src/lib.rs:1091:13
   8: main
             at ./main_debug_crash.cpp:162:18
   9: <unknown>
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
fatal runtime error: failed to initiate panic, error 5
[1]    36799 abort      RUST_BACKTRACE=1 ./yrs-ffi-tests`

xyzhangfred:
Here is the code I am referring to
![image](https://github.com/y-crdt/y-crdt/assets/44180807/98559eac-f608-4c5d-84d2-c72c86996fb2)

Horusiath:
[...]

PR #399, opened 2024-03-18
Handle the case where a split is about to happen in the middle of a UTF-16 surrogate pair

xyzhangfred:
I've identified an issue in the split_str function where splitting a string at an offset that divides a surrogate pair results in incorrect behavior. Specifically, when split_str is called with an offset that bisects a surrogate pair, the entire character is retained in the left part of the split, leaving the right part empty. This can lead to inaccuracies in length calculations ( there are places in the code where the "offset" is taken to be the correct length of the left splice, after the split) and potential empty strings where they're not expected.

Example:

Input: split_str("🌉", 1, OffsetKind::Utf16)
Current Output: left = "🌉", right = "" 
Expected Behavior: Ideally, the function should either somehow handle surrogate pairs gracefully (although I am not sure if that's possible with rust),or handle it in a way that does not lead to incorrect string lengths/ block splice lengths.

Temporary Workaround:
I've implemented a temporary workaround that replaces problematic splits with empty strings. This approach prevents crashes and highlights the issue, though it's far from an ideal solution.

This workaround is intended as a stopgap measure. I welcome suggestions for a more elegant and robust solution to this problem. Please feel free to discuss this further or propose alternatives.

### DIFF
--- a/yrs/src/block.rs
+++ b/yrs/src/block.rs
@@ -1477,8 +1477,8 @@ impl Deref for SplittableString {
     }
 }
 
-pub(crate) fn split_str(str: &str, offset: usize, kind: OffsetKind) -> (&str, &str) {
-    fn map_utf16_offset(str: &str, offset: u32) -> u32 {
+pub(crate) fn split_str(str: &str, offset: usize, kind: OffsetKind) -> Option<(&str, &str)> {
+    fn map_utf16_offset(str: &str, offset: u32) -> (usize, bool) {
         let mut off = 0;
         let mut i = 0;
         for c in str.chars() {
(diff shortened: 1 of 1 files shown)


xyzhangfred:
for reference: https://github.com/y-crdt/y-crdt/issues/386

Horusiath:
I don't think that's the root issue, but nonetheless the fact that you've found a repeatable case is very helpful. I'll take a look at it during Thu-Fri. The fix will arrive as part of v0.18.1 release by the end of this week.

Horusiath:
@xyzhangfred wouldn't you mind attaching the test case to recreate the issue? What I did was:

```rust
let doc = Doc::with_options(Options {
    client_id: 1,
    // use UTF-16, since this is the offset kind that Yjs uses (use od UTF16 encoding in JavaScript)
    offset_kind: OffsetKind::Utf16, 
    ..Default::default()
});
let txt = doc.get_or_insert_text("quill");
let str = "🇿🇿🇿🇿🇩🇩🇩🇿🇩🇩🇩🇩🇩🇿🇩🇩🇿🇩🇿🇩";
println!("length: {}", str.len());
txt.insert(&mut txn, 0, str);
txt.remove_range(&mut txn, 0, 5);
```

However this test doesn't bring any errors.

xyzhangfred:
Maybe it only happens in the C FFI? I used the following code to reproduce a crash (replacing main.cpp in y-crdt/tests-ffi, compile and run).

`#include <stdio.h>
#include <string.h>

extern "C" {
    #include "include/libyrs.h"
};

int main(){

    YDoc* doc = ydoc_new();
    Branch* txt = ytext(doc, "quill");
    YTransaction* txn = ydoc_write_transaction(doc, 0, NULL);
    ytext_insert(txt, txn, 0, "🇿🇿🇿🇿🇩🇩🇩🇿🇩🇩🇩🇩🇩🇿🇩🇩🇿🇩🇿🇩", NULL);
    ytext_remove_range(txt, txn, 0, 5); 

}`

Horusiath:
After checking out this code more thoroughly, I see that your example couldn't work in the first place when using using the default OffsetKind (which targets UTF8 byte lenghts per character). Deleting byte-range of 0..5, when affected characters are 4 bytes long, means splitting a character in half which is not allowed operation in Rust. 

Changing OffsetKind to Utf16 (so the same lengths as Yjs/JavaScript uses) fixes the issue.

PR #419, opened 2024-04-26
[yrs] fix `has_subscribers` on observer

### LINKED ISSUE #418: [yrs] `has_subscribers` on observer always return false
The `has_subscriber` method on the `Observer` always returns `false`.

```rs
use yrs::Observer;

fn main() {
    let observer: Observer<String> = Observer::new();

    assert!(!observer.has_subscribers());

    let _sub = observer.subscribe(move |_txn, _e| {});

    // Assertion fails
    assert!(observer.has_subscribers());
}
```

This breaks the awareness logic.

Edit: I've opened a PR that fixes this: #419

zignis:
Fixes the `has_subscribers` method. I've also added a test case for the same.

Fixes #418

### DIFF
--- a/yrs/benches/benches.rs
+++ b/yrs/benches/benches.rs
@@ -4,9 +4,9 @@ use rand::prelude::StdRng;
 use rand::{Rng, RngCore, SeedableRng};
 use std::cell::Cell;
 use std::collections::HashMap;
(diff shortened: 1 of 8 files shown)


Horusiath (approved):
Good catch 👍

PR #443, opened 2024-05-27
Compiled out time from the awareness MetaClientState

### LINKED ISSUE #442: Configurable time provider support Awareness
Time is not supported for the `wasm32-unknown-unknown` target, leading to a runtime panic. For such platforms would it be possible to add an option that takes in a time provider? This would allow awareness to be supported on all platforms.

ncthbrt:
Fixes #442. I'm not sure if this is an appropriate fix, but the alternative seemed to be to change the signature or to add a dependency on something like web_time which would drag in js_sys and so on.


Horusiath:
This was realized as part of #446

PR #444, opened 2024-06-05
rlib crate type for yffi

dileping:
Hi guys!

This PR allows `yffi` to be built as a Rust library.

Linking several Rust-produced static libs to a single executable is problematic, so it has to be done via an umbrella crate that reexports the APIs and is linked as a big single static lib. Enabling `rlib` for `yffi` allows to do just that.

This PR should not affect any other functionality.

### DIFF
--- a/yffi/Cargo.toml
+++ b/yffi/Cargo.toml
@@ -15,5 +15,5 @@ description = "Bindings for the Yrs native C foreign function interface"
 yrs = { path = "../yrs", version = "0.18.8", features = ["weak"] }
 
 [lib]
-crate-type = ["staticlib", "cdylib"]
+crate-type = ["rlib", "staticlib", "cdylib"]
 name = "yrs"


Horusiath (approved):
Thanks :) I'm happy to include it.

dileping:
@Horusiath Thanks for the fast reply!

I've just checked the logs for the failing tests. Apparently, the change that needs to be done is a little more significant than the minimal oneliner I hoped for. `yffi` exports its `staticlib` and `cdylib` named `yrs`, thus if I add `rlib`, it has a name conflict with the actual `yrs` crate. There are some different ways to address it:

1. Preserving current `yrs` naming for `yffi` crate exports: make current `yffi` crate an intermediate crate by making it `rlib` only and add another crate `yffi-exports` (maybe a better name) that will export `staticlib` and `cdylib` preserving the current `yrs` naming.
2. Almost identical to option 1, but keep the current `yffi` as the one that exports the C libraries and creates an intermediate that is an `rlib`.
3. Change `lib.name` in `yffi` from `yrs` to `yffi`. As a consequence, everything that depends on these C libraries will have to change their linker flags (which is not ideal).

I'd suggest implementing option 1 or 2 as the implications of that are minimal. I'd choose option 1. Would this work for you?

dileping:
@Horusiath  I did a little more digging into it and found out that `ywasm` actually does what I described in option 3 (exports itself as `ywasm` and not `yrs`):

```toml
[lib]
crate-type = ["cdylib", "rlib"]
```

I look forward to your opinion on what I should do with this PR. If you find option 3 OK (considering dependent libraries needing an update), I'd say it's a better long-term choice, IMO. 

Anyhow, please let me know what you think, and I prepare a PR according to your preference.

Horusiath:
@dileping I think that option 2 would be the best one out of three: AFAIK everyone who use static/dynamic libs for C interop, use yffi header files anyway. 

Another option would be to merge yffi into yrs as a dedicated module that can be turned on/off via eg. feature flag.

dileping:
@Horusiath, thanks for getting back to me on this. It makes sense. I'll try to arrange some time next week to move this forward.

dileping:
@Horusiath, I've just pushed the update here. Please check if that's okay with you. Also, if the PR is accepted, it would be great if you could publish `yffi-impl` on crates.io.

Issue #447, opened 2024-06-18
Error occurs in yffi when inserting nested Maps into document


twaddell:
Inserting a more complex nested map into a document appears to work fine in yrs but fails in yffi.
I am running on a Windows environment.

The following test case in yffi fails
```
TEST_CASE("YMap multiple nested maps") {
    YDoc* doc = ydoc_new_with_id(1);
    Branch* map = ymap(doc, "map");
    YTransaction* txn = ydoc_write_transaction(doc, 0, NULL);

    char* key = (char*)"text";
    YInput value = yinput_string("Nested data");
    YInput innerMap = yinput_ymap(&key, &value, 1);

    char* key2 = (char*)"innerMap";
    YInput outerMap = yinput_ymap(&key2, &innerMap, 1);

    ymap_insert(map, txn, "outerMap", &outerMap);
    int length = ymap_len(map, txn);
    ytransaction_commit(txn);
    REQUIRE(length == 1);

    ydoc_destroy(doc);
}
```

The following error is shown in the console when this test case is run:
```
thread '<unnamed>' panicked at yffi\src\lib.rs:2488:17:
Unrecognized YVal value tag.
```

Simplifying the nesting in yffi will allow the test to pass
```
TEST_CASE("YMap nested map") {
    YDoc* doc = ydoc_new_with_id(1);
    Branch* map = ymap(doc, "map");
    YTransaction* txn = ydoc_write_transaction(doc, 0, NULL);

    char* key = (char*)"text";
    YInput value = yinput_string("Nested data");
    YInput innerMap = yinput_ymap(&key, &value, 1);

    ymap_insert(map, txn, "innerMap", &innerMap);
    int length = ymap_len(map, txn);
    ytransaction_commit(txn);
    REQUIRE(length == 1);

    ydoc_destroy(doc);
}
```

The following test case in yrs passes
```
  #[test]
  fn map_nested() {
      let d1 = Doc::with_client_id(1);
      let m1 = d1.get_or_insert_map("map");
      let mut t1 = d1.transact_mut();

      let mut innerMap: HashMap<String, _> = HashMap::new();
      innerMap.insert("text".to_owned(), "Nested data");
      
      let mut outerMap: HashMap<String, _> = HashMap::new();
      outerMap.insert("innerMap".to_owned(), MapPrelim::from(innerMap));

      m1.insert(&mut t1, "outerMap".to_owned(), MapPrelim::from(outerMap));
      
[...]

PR #452, opened 2024-06-25
yffi: stop forcing Any conversion on Y_MAP input

Horusiath:
Related #447

### DIFF
--- a/yffi/src/lib.rs
+++ b/yffi/src/lib.rs
@@ -5,7 +5,7 @@ use std::ops::{Deref, RangeBounds};
 use std::ptr::{null, null_mut};
 use std::sync::atomic::{AtomicPtr, Ordering};
 use std::sync::Arc;
-use yrs::block::{ClientID, ItemContent, Prelim, Unused};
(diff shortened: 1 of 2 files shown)


yffi/src/lib.rs
```diff
@@ -2562,7 +2571,7 @@ impl Prelim for YInput {
                             .to_str()
                             .unwrap()
                             .to_owned();
-                        let value = values.offset(i).read().into();
```
Horusiath, line 2565: Apparently the issue lied here: when we try to insert prelim map filled with initial content, we tried to forcefully convert all nested values onto Any type, even if they were prelim types themselves.

yffi/src/lib.rs
```diff
@@ -2177,7 +2177,7 @@ pub unsafe extern "C" fn yxmltext_insert_embed(
         .expect("provided transaction was not writeable");
     let txt = XmlTextRef::from_raw_branch(txt);
     let index = index as u32;
-    let content: Any = content.read().into();
```
Horusiath, line 2180: The same issue existed in inserting embed elements into text: we couldn't insert other Y types as text embeds.

PR #454, opened 2024-06-28
Fixing FFI Unit Tests for Windows

skeptopic:
This PR fixes the FFI unit tests on Windows. 

Thanks to @MarkJGx (https://github.com/MarkJGx) for helping do this.

CMake:
- Ensure C++20
- Ensure UTF-8 Encoding
C++:
- Ensure YRS destroy functions are called where needed
- explicitly allocate char*s

### DIFF
--- a/tests-ffi/CMakeLists.txt
+++ b/tests-ffi/CMakeLists.txt
@@ -1,14 +1,16 @@
 cmake_minimum_required(VERSION 3.15.3)
 project(yrs-ffi-tests)
-set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 message(STATUS "current project dir: " ${PROJECT_SOURCE_DIR})
 add_executable(yrs-ffi-tests main.cpp)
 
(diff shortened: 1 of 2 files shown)


Horusiath:
Thanks for your contribution. Bless you guys.

Horusiath (commented):
If I remember correctly one of the issues with interop between Rust and C++ was that Rust used different allocator calls between different OSes. 

I don't remember exactly - it was many months ago - but I think there was an issue related to `free` calls done inside of observer callbacks, which blowed up on Windows because Rust-allocated that should be released using `HeapFree` there instead.

skeptopic:
> If I remember correctly one of the issues with interop between Rust and C++ was that Rust used different allocator calls between different OSes.
> 
> I don't remember exactly - it was many months ago - but I think there was an issue related to `free` calls done inside of observer callbacks, which blowed up on Windows because Rust-allocated that should be released using `HeapFree` there instead.

Our read on this is that the pointer was allocated by Rust so it should be also de-allocated by Rust as well. 

Thus by destroying that memory through `ybinary_destroy` we ensure the same allocation/deallocation strategy for the memory no matter which platform

Horusiath (approved):

PR #468, opened 2024-07-08
UndoManager: undo_stack and redo_stack access methods

### LINKED ISSUE #466: Undo manager undo/redo stacks
AFAIU the undo manager's `undo_stack` and `redo_stack` are not accessible in the public API, unlike in Yjs.
I have a case where I need to know if there is only one item left to undo, but I don't want to undo it. It seems that having access to the `undo_stack`'s length would be a solution.

Horusiath:
Closes #466


yrs/src/undo.rs
```diff
@@ -35,7 +35,6 @@ use crate::{DeleteSet, Doc, Observer, Transact, TransactionMut, ID};
 /// - [UndoManager::observe_item_popped], which is fired whenever [StackItem] is being from undo
 ///    manager as a result of calling either [UndoManager::undo] or [UndoManager::redo] method.
 #[repr(transparent)]
-#[derive(Clone)]
```
Horusiath, line 38: Actually, relying on derived clone method here for Undo will break all `UndoManager` methods using `&mut self`.

PR #470, opened 2024-07-08
Remove stale comment about Subscriptionid

kyle-leonhard:
SubscriptionId was removed some time ago, and, as far as I can tell, there isn't functionality to replace it. I stumbled on this until I realized I needed to turn on the "sync" feature to pass a subscription between threads.


Horusiath (approved):

Horusiath:
Thanks. It would be probably a good idea to point out in the documentation comments that `Subscription` now requires a `sync` feature flag to implement Send+Sync itself.

PR #481, opened 2024-08-18
Async transactions

Horusiath:
This PR introduces serveral changes:
1. `Transact::transact` and `Transact::transact_mut` methods now won't panic when transaction couldn't be acquired. Instead they will block a thread. You can always simulate old behaviour by using `doc.try_transact().unwrap()`.
2. Introduces new trait `AsyncTransact` which has API similar to `Transact`, however it's versions of `transact` and `transact_mut` returns a Future that can we awaited within async functions.
3. Removed `Doc::options` since internally we had to wrap options in some dedicated arc-swap types (for subdocument options override behaviour). Instead `Doc::auto_load`/`Doc::should_load`/`Doc::collection_id` are now dedicated methods instead of being just fields of `Doc::options`.
4. `Doc::collection_id` type changed from `String` &rarr; `Arc<str>`.

## Cargo criterion benchmarks

Changes make some of the tests improve/regress within 12% threshold - I'd say this is a fair tradeoff and since the tests are very short they could be influenced by other factors (even my comp temperature).
```
[B1.1] Append N characters/6000
                        time:   [131.78 ms 132.50 ms 133.35 ms]
                        change: [-2.6588% -1.6267% -0.6511%] (p = 0.01 < 0.05)
                        Change within noise threshold.

[B1.2] Insert string of length N/1
                        time:   [5.4072 µs 5.4103 µs 5.4140 µs]
                        change: [-0.5335% +0.8567% +2.2592%] (p = 0.29 > 0.05)
                        No change in performance detected.

[B1.3] Prepend N characters/6000
                        time:   [5.5093 ms 5.5892 ms 5.6481 ms]
                        change: [+12.560% +14.283% +15.924%] (p = 0.00 < 0.05)
                        Performance has regressed.

[B1.4] Insert N characters at random positions/6000
                        time:   [43.529 ms 43.634 ms 43.727 ms]
                        change: [+2.1378% +2.4059% +2.6693%] (p = 0.00 < 0.05)
                        Performance has regressed.

[B1.5] Insert N words at random positions/6000
                        time:   [23.627 ms 23.827 ms 23.946 ms]
                        change: [-0.7381% -0.0743% +0.5748%] (p = 0.84 > 0.05)
                        No change in performance detected.

Benchmarking [B1.7] Insert/Delete strings at random positions/6000: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 8.9s or enable flat sampling.
[...]

PR #488, opened 2024-08-26
add tests for transact in multi-threaded behaviour

Horusiath:


yrs/src/transact.rs
```diff
@@ -233,3 +233,44 @@ pub enum TransactionAcqError {
...
+        const N: usize = 3;
+        let barrier = Arc::new(Barrier::new(N + 1));
+
+        let start = Instant::now();
+        for i in 0..N {
+            let d = doc.clone();
+            let t = txt.clone();
+            let b = barrier.clone();
+            std::thread::spawn(move || {
+                // let mut txn = d.try_transact_mut().unwrap(); // this will hang forever
```
davidbrochart, line 258: From [this comment](https://github.com/y-crdt/y-crdt/pull/481#issue-2472045080):
> 1. `Transact::transact` and `Transact::transact_mut` methods now won't panic when transaction couldn't be acquired. Instead they will block a thread. You can always simulate old behaviour by using `doc.try_transact().unwrap()`.

I understood that `d.try_transact_mut().unwrap()` will panic (for all other threads after the first one acquired the transaction), not that it will hang forever?

Horusiath: In this case it will hang forever, since panic inside of the thread will prevent execution of `barrier.wait()`, which means that we won't be able to pass over the barrier on the main test thread.

davidbrochart (approved):
Thanks!

PR #489, opened 2024-08-27
Undo manager transactions

Horusiath:
This PR applies several changes:

1. `UndoManager` undo/redo methods are now split to resemble new async capabilities of document store: 
    - `try_undo`/`try_redo` will try to obtain underlying document store transaction and return error on failure.
    - `undo_blocking`/`redo_blocking` will obtain transactions, waiting by blocking the thread if necessary.
    - `undo`/`redo` now returns futures that will obtain transactions asynchronously.
2. `yundo_manager_can_undo` and `yundo_manager_can_redo` are now replaced by `yundo_manager_undo_stack_len` and `yundo_manager_redo_stack_len`.

### DIFF
--- a/yrs/src/undo.rs
+++ b/yrs/src/undo.rs
@@ -33,8 +33,10 @@ use crate::{DeleteSet, Doc, Observer, Transact, TransactionAcqError, Transaction
 ///    item finished.
 /// - [UndoManager::observe_item_popped], which is fired whenever [StackItem] is being from undo
 ///    manager as a result of calling either [UndoManager::undo] or [UndoManager::redo] method.
-#[repr(transparent)]
-pub struct UndoManager<M>(Arc<Inner<M>>);
+pub struct UndoManager<M> {
+    state: Arc<Inner<M>>,
+    doc: Doc,
+}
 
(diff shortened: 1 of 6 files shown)


yrs/src/undo.rs
```diff
@@ -479,29 +477,35 @@ where
...
-        let mut txn = inner.doc.try_transact_mut()?;
+    ///
+    /// # Deadlocks
+    ///
+    /// In order to perform its function, this method must guarantee that underlying document store
+    /// is not being modified by another running `TransactionMut`. It does so by acquiring
+    /// a read-only transaction itself. If transaction couldn't be acquired (because another
+    /// read-write transaction is in progress), it will hold current thread until, acquisition is
+    /// available.
+    pub fn clear(&mut self) {
```
davidbrochart, line 493: It looks like there is no `try_clear()`, as you did for other methods?

Horusiath: Yes, I'll need to add that in subsequent release. As of now, it will block current thread but only if there's a pending read-write transaction (since clear only requires a readonly access.

PR #511, opened 2024-11-07
DeleteSet: optimize GC squash for improved performance

MarkJGx:
This change set introduces an optimization to Yrs garbage collection squashing.

# Details

To further enhance the scalability of the garbage collection (GC) process in DeleteSet, we've introduced a bulk squashing method that optimizes performance when handling large quantities of GC block cells. The previous per-block handling of squash calls worked well for many use cases. By implementing bulk squashing, we aim to reduce the overhead from constant reallocations while removing individual BlockCells (both GC and Block types).

# Results

In our internal benchmark focusing on contiguous removals, we observed that the delete operation GC collection time decreased from 118 seconds to 400 milliseconds in release builds, representing a substantial performance improvement. (Please note, the benchmark uses proprietary types and cannot be shared publicly at this time.)

### DIFF
--- a/yrs/src/block_store.rs
+++ b/yrs/src/block_store.rs
@@ -5,15 +5,20 @@ use crate::types::TypePtr;
 use crate::utils::client_hasher::ClientHasher;
 use crate::*;
 use std::collections::hash_map::Entry;
-use std::collections::HashMap;
+use std::collections::{HashMap};
 use std::hash::BuildHasherDefault;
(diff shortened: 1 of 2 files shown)


Horusiath:
Thanks @MarkJGx . That sounds fantastic :) I'll review (and hopefully merge) it over the weekend.

Horusiath (approved):

PR #524, opened 2025-02-11
Derive Debug on Yrs Events and all dependent structs

shadanan:
Allows debug printing of the Event and Events object which is useful during development. For example:

```rust
let doc = Doc::new();
let my_map = doc.get_or_insert_map("my_map");
my_map.observe_deep(|txn, events| println!("Handling deep my_map events: {events:?}"));
```

This enables logging information about the kinds of events, for example:

Handling deep my_map events: Events([Map(MapEvent { current_target: 'my_map', target: MapRef(<1278253509#0>), keys: UnsafeCell { .. } })])

Handling deep my_map events: Events([Map(MapEvent { current_target: 'my_map', target: MapRef('my_map'), keys: UnsafeCell { .. } }), Array(ArrayEvent { current_target: 'my_map', target: ArrayRef(<948583607#1450>), change_set: UnsafeCell { .. } })])


Horusiath:
Tbh. I have some doubts about exposing Display/Debug info (of maybe hide if behind feature flag). The reason is that currently WASM target is ~300kB (compared to ~30kB of Yjs), and I'll eventually start trimming some of its parts.

PR #545, opened 2025-04-25
align Skip encoding with Yjs format

catssay:
Hey, thanks for the great work on Yjs — we’ve been able to add real-time collaboration to our product because of it!

We’ve recently been trying to migrate our Yjs server over to yrs. During testing, we hit a panic when loading updates into YDoc. After digging in, it looks like yrs isn’t able to decode updates created with encode_state_as_update_v2.

I spent a few hours comparing things between Yjs and Yrs and found that the encoding of Skip differs. After tweaking the code to match Yjs’s behavior, the binary output matched and the panic was gone.

I’ve attached the update file that was causing the issue (encoded with`encode_state_as_update_v1` and write it with `std::fs::write`), in case it helps.

Also, fair warning — I’m pretty new to Rust and still learning my way around Yjs/Yrs, I don't known how to write the tests, so let me know if anything looks off.

### DIFF
--- a/yrs/src/update.rs
+++ b/yrs/src/update.rs
@@ -944,7 +944,8 @@ impl BlockCarrier {
             }
             BlockCarrier::Skip(x) => {
                 encoder.write_info(BLOCK_SKIP_REF_NUMBER);
-                encoder.write_len(x.len - offset);
+                // write as VarUint because Skips can't make use of predictable length-encoding
+                encoder.write_var(x.len - offset);
             }
             BlockCarrier::GC(x) => {
                 encoder.write_info(BLOCK_GC_REF_NUMBER);


catssay:
> Hey, thanks for the great work on Yjs — we’ve been able to add real-time collaboration to our product because of it!
> 
> We’ve recently been trying to migrate our Yjs server over to yrs. During testing, we hit a panic when loading updates into YDoc. After digging in, it looks like yrs isn’t able to decode updates created with encode_state_as_update_v2.
> 
> I spent a few hours comparing things between Yjs and Yrs and found that the encoding of Skip differs. After tweaking the code to match Yjs’s behavior, the panic was gone.
> 
> I’ve attached the update file that was causing the issue (encoded with`encode_state_as_update_v1` and write it with `std::fs::write`), in case it helps.
> 
> Also, fair warning — I’m pretty new to Rust and still learning my way around Yjs/Yrs, I don't known how to write the tests, so let me know if anything looks off.

The binary is not full matched, there is still one byte not matched, but encoding and decoding with v2 are fine, I may figure out later.

Horusiath:
Thanks for your fix. Are you able to provide a test case that shows, it indeed solves the issue?

yrs/src/update.rs
```diff
@@ -944,7 +944,8 @@ impl BlockCarrier {
             }
             BlockCarrier::Skip(x) => {
                 encoder.write_info(BLOCK_SKIP_REF_NUMBER);
-                encoder.write_len(x.len - offset);
+                // write as VarUint because Skips can't make use of predictable length-encoding
+                encoder.write_var(x.len - offset);
```
Horusiath, line 948: I've tested your example, and as it turns out this line change alone is enough to make it pass - the change above doesn't seem to affect results. I've also prepared artificial test with skip blocks inside and it also passed with only this line present.

yrs/src/block.rs
```diff
@@ -443,11 +443,12 @@ impl ItemPtr {
             let item = self.deref_mut();
             let client = item.id.client;
             let clock = item.id.clock;
+            let len = item.len;
             let content = item.content.splice(offset as usize, encoding).unwrap();
             item.len = offset;
             let mut new = Box::new(Item {
                 id: ID::new(client, clock + offset),
-                len: content.len(OffsetKind::Utf16),
+                len: len - offset,
```
Horusiath, line 451: I'm not sure if this change is necessary whatsoever.

Horusiath (changes requested):
Looks like your fix is ok, but apparently this fix seems to be a one liner.

yrs/src/block.rs
okxiaoliang4: This change is for performance. Avoid multiple calculations of string length to affect efficiency.

okxiaoliang4: > I'm not sure if this change is necessary whatsoever.

@Horusiath This change is for performance. Avoid multiple calculations of string length to affect efficiency.

okxiaoliang4, line 451:
```suggestion
                len: item.len - offset,
```

okxiaoliang4 (changes requested):

yrs/src/block.rs
Horusiath: I'll be happy to add this change then. Can you rebase your changes please?

okxiaoliang4: Ok. @catssay will handle this.

3yellowbear: Thanks @Horusiath and @okxiaoliang4 for your reply, I've rebased changes.

Horusiath (approved):

PR #563, opened 2025-06-19
[WIP] Remove locks

Horusiath:
This PR is an ongoing work to refactor yrs to remove locks from the API and push towards classic Rust borrow-checker reinforced model. It's for few reasons:
1. To reduce the size of binary (necessary for WASM target).
2. Locks implemented on yrs level are useless when we need to interop with higher level languages, which need their own locking pattern anyway.

This PR aims to introduce a number of breaking changes:
- [x] Remove locks and replace them with immutable/mutable `Doc` refrences.
- [x] Transaction state for writeable transactions is moved into separate object, that can be initialized on demand.
- [x] Remove unsafe code and replace it with different access paterns that will affect the other parts of the API, like:
  - [x] UndoManager (undo/redo now require `&mut Doc` access).
  - [x] Event API (event's no longer require transaction to lazily evaluate some of their properties).
- [ ] Remove different transaction types: only a single `Transaction` kind with either immutable/mutable access to underlying `Doc`.
- [ ] Make `Doc` into a trait-like type: so that we can make `Transaction<D> where D: BorrowMut<Doc>` which will work together with i.e. `Rc<RefCell<Doc>>` (for ywasm) or `Arc<Mutex<Doc>>` (for multi-threaded Rust) borrowed references.
- [ ] Make `Awareness` into trait-like type: awareness should be composable into custom types.

### DIFF
--- a/yffi/src/lib.rs
+++ b/yffi/src/lib.rs
@@ -4541,14 +4541,14 @@ pub unsafe extern "C" fn yundo_manager_redo(mgr: *mut YUndoManager, doc: *mut Do
 #[no_mangle]
 pub unsafe extern "C" fn yundo_manager_undo_stack_len(mgr: *mut YUndoManager) -> u32 {
     let mgr = mgr.as_mut().unwrap();
-    mgr.undo_len() as u32
+    mgr.undo_stack().len() as u32
 }
 
 /// Returns number of elements stored on redo stack.
 #[no_mangle]
 pub unsafe extern "C" fn yundo_manager_redo_stack_len(mgr: *mut YUndoManager) -> u32 {
     let mgr = mgr.as_mut().unwrap();
-    mgr.redo_len() as u32
+    mgr.redo_stack().len() as u32
 }
 
 /// Subscribes a `callback` function pointer to a given undo manager event. This event will be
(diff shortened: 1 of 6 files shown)


axkibe:
I think this is an excellent idea, IMO there should be a core library that makes as few as possible assumptions on the runtime context (if its single threaded, a FFI, or multithreaded) and possibly one or more comfort wrappers around this, that may make mutexes etc.

I think it would be better, if there wouldn't be a read transaction at all, a given data tree would then be immutable for all time and can be passed around with Arc<> handles (or Rc<> in single thread RT) and to make changes a mutable copy is created which does keep the changes in a transaction struct. Once it's done it's frozen and be passed around as the new data tree.

This would also open up multi-threaded server runtime where the server could handle two clients at once, each with their change set and when they are done merge the two versions into one new..

Horusiath:
@axkibe the entire architecture of yjs/yrs relies on mutable object graph and we won't change that here. I'm experimenting on [y-crdt variant over LMDB](https://github.com/horusiath/ysr) (which is immutable copy-on-write persistent key-value store), but it won't have performance characteristics or some of the features that yrs has.

Issue #568, opened 2025-06-28
Xml setAttribute, insert_attribute, get_attribute should support full JSON values (not just strings)


planerist:
In the original [yjs](https://github.com/yjs/yjs) implementation, YXmlElement.setAttribute(name: string, value: string | Y.AbstractType) is documented to only accept strings or Y.AbstractType. However, in real-world usage, many libraries (e.g. slate-yjs) rely on being able to pass full JSON-serializable values — including boolean, number, null, etc.

## Problem in y-crdt:

In y-crdt, this compatibility is currently broken because non-string values passed to setAttribute are silently coerced to strings. This causes subtle and hard-to-debug breakages in downstream libraries like Slate-yjs, which assume correct preservation of types (e.g., false turning into "false" breaks logic relying on boolean checks).

## Why this matters:

- This silent coercion introduces bugs that are very hard to track down.

- Many libraries have organically adopted the broader value semantics of setAttribute (accepting all JSON primitives).

- Current behavior diverges from what the ecosystem expects, leading to broken integrations.

## Proposal:

Update y-crdt's YXmlElement.setAttribute to support full JSON values (string, number, boolean, null, etc.)—not just strings.

This would:

- Align y-crdt behavior with common practice and library expectations.

- Eliminate silent coercion bugs.

- Likely be easy to implement using the existing EmbedPrelim mechanism.

## Open question:

Should this change:

1. Modify the existing setAttribute method (breaking from the original yjs-style API)?

2. Introduce a parallel method (e.g. setJsonAttribute) to preserve backward compatibility?

Happy to contribute a PR depending on the preferred direction.

PR #569, opened 2025-06-29
Add support for full JSON values in XML attribute methods

planerist:
See #568 for context and discussion on JSON attribute support in XML methods.

yrs
 
- Added support for attributes with Any type via new get_attribute_any and attributes_any methods

 - Preserved existing attribute* methods for backward compatibility

 - Enhanced internal logic to support Any-typed attributes across XML elements

ywasm

 - Extended xml.attrs* to handle Any-typed attributes

 - Added test coverage for the new functionality


Horusiath:
I'll need a second to properly review it.

Issue #582, opened 2025-10-23
Pending delete set from a new client is silently dropped when applying updates out of order


procr1337:
It appears that [`TransactioMut::apply_delete`](https://github.com/y-crdt/y-crdt/blob/0c74f74ab71646b4ec1ef01bbc841a6187bc6342/yrs/src/transaction.rs#L607) discards delete set for clients that it hasn't seen before, leading to some weird edge cases where updates being applied out of order can lead to incorrect results. Example:

```rust
use yrs::{
    GetString as _, ReadTxn as _, StateVector, Text as _, Transact as _, Update,
    WriteTxn as _, updates::decoder::Decode,
};

fn copy_doc(doc: &yrs::Doc) -> yrs::Doc {
    let tx = doc.transact();
    let update = tx.encode_diff_v2(&StateVector::default());
    drop(tx);

    let new_doc = yrs::Doc::new();
    let mut new_tx = new_doc.transact_mut();
    new_tx
        .apply_update(Update::decode_v2(&update).unwrap())
        .unwrap();
    drop(new_tx);

    new_doc
}

fn main() {
    let doc = yrs::Doc::new();

    let (upd1, upd2) = {
        let doc2 = copy_doc(&doc);
        let mut tx = doc2.transact_mut();
        let text = tx.get_or_insert_text("example");
        text.insert(&mut tx, 0, "foo");
        let upd1 = tx.encode_update_v2();
        drop(tx);

        let mut tx = doc2.transact_mut();
        let text = tx.get_or_insert_text("example");
        text.remove_range(&mut tx, 0, 1);
        let upd2 = tx.encode_update_v2();
        assert_eq!(text.get_string(&tx), "oo");
        drop(tx);

        (upd1, upd2)
    };

    let doc3 = copy_doc(&doc);
    let mut tx = doc3.transact_mut();
    tx.apply_update(Update::decode_v2(&upd2).unwrap()).unwrap();
    tx.apply_update(Update::decode_v2(&upd1).unwrap()).unwrap();
    let text = tx.get_or_insert_text("example");
    assert_eq!(text.get_string(&tx), "oo");
}
```

This fails with

```rust
thread 'main' panicked at .../pending_deletes.rs:47:5:
assertion `left == right` failed
  left: "foo"
 right: "oo"
```

From my limited understanding (and from all test cases passing), it would be sufficient to handle the unseen client case as follows:

```rust
[...]

PR #583, opened 2025-10-23
Fix for pending/unapplied deletes from a previously unseen client being discarded

procr1337:
See https://github.com/y-crdt/y-crdt/issues/582


Horusiath (approved):
Thanks for spotting that out!

PR #594, opened 2026-01-19
refactor(ywasm): replace gloo-utils with serde-wasm-bindgen

satoren:
Replace gloo-utils (which uses serde_json internally) with serde-wasm-bindgen for more efficient JS<->Rust serialization without JSON intermediate format.

Key changes:
- Replace gloo-utils dependency with serde-wasm-bindgen
- Add to_js() helper using Serializer::json_compatible() to ensure maps serialize as plain JS objects (not ES2015 Map)
- Update all JsValue::from_serde/into_serde calls to use the new APIs

This reduces the optimized WASM binary size by ~78KB (~9% reduction):
- Before: 837KB (with wasm-opt + strip)
- After:  759KB (with wasm-opt + strip)

All 89 tests pass.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -388,19 +388,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "gloo-utils"
-version = "0.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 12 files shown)


Horusiath (approved):
Thanks! Looks great :)

PR #603, opened 2026-03-30
fix: treat logically-deleted items as unknown parent in repair()

LucasXu0:
A logically-deleted item (is_deleted flag set) cannot serve as a parent branch. yjs handles this by evaluating content.type, which returns undefined for non-Type items (e.g. Format), silently dropping the child. yrs only checked ItemContent::Deleted (GC state) and missed the is_deleted() flag, causing a spurious InvalidParent error on documents containing deleted Format items as parent refs.

### DIFF
--- a/yrs/src/block.rs
+++ b/yrs/src/block.rs
@@ -1325,16 +1325,25 @@ impl Item {
             TypePtr::ID(id) => {
                 let ptr = store.blocks.get_item(id);
                 if let Some(item) = ptr {
-                    match &item.content {
-                        ItemContent::Type(branch) => {
-                            TypePtr::Branch(BranchPtr::from(branch.as_ref()))
-                        }
-                        ItemContent::Deleted(_) => TypePtr::Unknown,
-                        other => {
-                            return Err(UpdateError::InvalidParent(
-                                id.clone(),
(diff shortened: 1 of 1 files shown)


Horusiath:
I'm looking at yjs code equivalent and it doesn't work exactly the way you say it does: https://github.com/yjs/yjs/blob/612fef6c69eb44676019bd785128515c88467f7a/src/utils/encoding.js#L656-L662

Sure, if item content is not `ContentType`, then the `struct.parent` will be undefined, but that has nothing to do with deleted flag check.

LucasXu0:
I was wrong about the yjs behavior earlier. In yjs, `struct.parent` ends up undefined because `content.type` returns undefined for non-ContentType items (like ContentFormat) — not for the reason I said.

The actual problem I'm running into: there's a bin that yjs decodes fine, but yrs chokes on it with an InvalidParent error. `InvalidParent(ID { client: 2596724816, clock: 5251 }, 6)`

Horusiath:
Can you provide an update that can be used to reproduce this problem? I mean one that passes through Yjs, but fails in Yrs? It would be good as a confirmation of the bug being fixed and for the possible regressions.

PR #604, opened 2026-04-03
fix: read DeleteSet client IDs as u64 (Yjs v14 compat)

nf-matt:
## Problem

Yjs v14 changed client ID generation from `random.uint32` to `random.uint53`. The `IdSet` decoder in yrs reads client IDs as `u32`, truncating values >= 2^32.

This causes `merge_updates_v2` to produce merged updates with wrong DeleteSet client IDs — the DS references a truncated value that doesn't match any client in the state vector.

## Reproduction

```rust
let doc_a = Doc::with_options(Options {
    client_id: 4_165_229_750_201_522, // typical Yjs v14 ID (> u32::MAX)
    ..Default::default()
});
// create state, overwrite a field in doc_b (different large client ID)
// merge updates → DS client ID is silently truncated
```

Regression test included.

## Fix

One-line change in `id_set.rs:529`: `let client: u32` → `let client: u64`.

The encoder already writes full u64 via `encoder.write_var(client_id)` where `ClientID = u64`, so only the decode side needed fixing.

## V1 decoder (separate issue)

`DecoderV1::read_id` and `read_client` have the same `u32` pattern. Fixing those is a larger change — it affects CRDT conflict resolution ordering (client IDs are used for tie-breaking), so it needs matching changes deeper in the integration layer. Happy to follow up separately.

---

Update: I just noticed closed PR #590 that wasn't merged, which is directly related to this.

### DIFF
--- a/yrs/src/id_set.rs
+++ b/yrs/src/id_set.rs
@@ -526,7 +526,7 @@ impl Decode for IdSet {
         let mut i = 0;
         while i < client_len {
             decoder.reset_ds_cur_val();
-            let client: u32 = decoder.read_var()?;
+            let client: u64 = decoder.read_var()?;
             let range = IdRange::decode(decoder)?;
             set.0.insert(client as ClientID, range);
             i += 1;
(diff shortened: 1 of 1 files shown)


Horusiath:
Fixed as part of #612 . We also needed to keep old behaviour around for sake of compatibility with old updates (see: `small-client` feature flag and `pending_update_integration` test).

PR #606, opened 2026-04-06
fix delete set application after collected history

### LINKED ISSUE #605: Applying Yjs Updates to a Yrs Document Produces Incorrect XML
When applying two updates exported by Yjs to a fresh Yrs document, Yrs produces a different final XML tree than Yjs.

A summary of the reproduction flow:

1. Use two Yjs clients to build a small XML history under `Y.XmlFragment`.
2. Let one client set an attribute on a temporary XML node, then delete that node, so the history includes a deleted cross-client XML segment.
3. Build the visible tree as `<p><a></a></p>` and export it as the `before` update.
4. Load that state into another Yjs document, replace `<a>` with `<b>`, and export the missing state as a state-vector diff.
5. In Yjs, applying `before` followed by `diff` produces `<p><b></b></p>`.
6. In Yrs 0.25.0, applying the same two updates produces `<p><b></b><a></a></p>`.

I put together a minimal repro here:
https://github.com/mslxl/yrs-xml-syncstep2-repro

And here is the CI run showing the result:
https://github.com/mslxl/yrs-xml-syncstep2-repro/actions/runs/24019117000/job/70044203875

mslxl:
Fixes #605, the fix results can be seen in the [fix branch](https://github.com/mslxl/yrs-xml-syncstep2-repro/tree/fix)([CI Result](https://github.com/mslxl/yrs-xml-syncstep2-repro/actions/runs/24029662293/job/70075547946)) of the minimal reproduction repo.

This change fixes `DeleteSet` application when the start of a delete range falls inside collected history and the next live item still needs to be processed.

I am not very familiar with Yrs/Yjs, so while this change fixes the reproduced cases, I am not sure if any break changes exist. But all unit tests are passed. Please review carefully, thanks.

### DIFF
--- a/yrs/src/transaction.rs
+++ b/yrs/src/transaction.rs
@@ -620,7 +620,18 @@ impl<'doc> TransactionMut<'doc> {
                         }
                         // We can ignore the case of GC and Delete structs, because we are going to skip them
                         if let Some(mut index) = blocks.find_pivot(clock) {
-                            // We can ignore the case of GC and Delete structs, because we are going to skip them
+                            // The delete range may start in collected history.
+                            // Advance to the next materialized item so the remainder of the range
+                            // is still applied to any later live items.
+                            while index < blocks.len() && blocks[index].as_item().is_none() {
+                                index += 1;
+                            }
+
+                            if index == blocks.len() {
+                                continue;
+                            }
+
+                            // We can ignore the case of Delete structs, because we are going to skip them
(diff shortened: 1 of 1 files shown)


Horusiath:
Thanks for the reproducible scenario. While your solution fixes this one case, I'm not sure if it will fix others. I've added #614 which in terms of behaviour is more aligned with yjs and hopefully fixes the problem. I'll close this one.

PR #610, opened 2026-04-26
DecoderV1: read clientIDs as u64 (fixes silent truncation above 2^32)

### LINKED ISSUE #609: DecoderV1::read_client and read_id silently truncate clientIDs above 2^32 (lossless interop with yjs JS peer broken)
### Summary

`yrs::updates::decoder::DecoderV1` reads client IDs as `u32` in two places, then casts to `ClientID = u64`. Because `read_var<u32>` uses `u32::wrapping_shl` per byte, any client ID whose varuint encoding requires more than 32 bits has its upper bits silently dropped (XOR-rotated into the lower 32 bits). The encoder side is correct (writes full `u64` via `write_var<u64>`); the wire format is correct (`lib0`'s `readVarUint` reads up to 53 bits losslessly via JS `Number`); only the V1 decoder is buggy.

`DecoderV2` reads client IDs as `u64` and is unaffected.

### Two affected sites in `src/updates/decoder.rs`

1. **`DecoderV1::read_client`** (line 143)

   ```rust
   fn read_client(&mut self) -> Result<ClientID, Error> {
       let client: u32 = self.cursor.read_var()?;
       Ok(client as ClientID)
   }
   ```

   Reads the per-client header in the v1 update binary (one entry per writer in `Update::blocks.clients`).

2. **`DecoderV1::read_id`** (lines 85-88)

   ```rust
   fn read_id(&mut self) -> Result<ID, Error> {
       let client: u32 = self.read_var()?;
       let clock = self.read_var()?;
       Ok(ID::new(client as ClientID, clock))
   }
   ```

   Reads block-origin (`left_id` / `right_id`) clientIDs — i.e., the IDs of left/right neighbour items that determine CRDT ordering. **More dangerous than `read_client`**: an op whose left/right origin references a remote writer with `client > 2^32` silently re-targets a different (truncated) ancestor on decode, which breaks Yjs's CRDT convergence semantics.

### Reproduction

`yrs = "0.21.3"`, also confirmed on `0.25.0` and `main`:

```rust
use yrs::{Doc, Options, OffsetKind, Text, Transact, ReadTxn, StateVector};
use yrs::updates::decoder::Decode;

fn main() {
    let cid: u64 = 7_079_134_143_100_776; // 0x19266f5beab768, ≈ 7×10^15, < 2^53
    let doc = Doc::with_options(Options {
        client_id: cid,
        skip_gc: false,
        offset_kind: OffsetKind::Bytes,
        ..Default::default()
    });
    {
        let text = txn.get_or_insert_text("content");
        text.insert(&mut txn, 0, "x");
    }
    let bytes = doc.transact().encode_state_as_update_v1(&StateVector::default());
    let update = yrs::Update::decode_v1(&bytes).unwrap();
    let decoded: Vec<u64> = update.state_vector().iter().map(|(c, _)| *c).collect();
    println!("encoded: {cid} (0x{cid:x})");
    println!("decoded: {decoded:?}"); // [1543223144] (0x5bfbb768) — lower 32 bits XOR-rotated
    assert_eq!(decoded, vec![cid], "expected lossless round-trip");
}
```

The same bytes parsed by JS `yjs` via `Y.parseUpdateMeta` return the original `7079134143100776` losslessly (`lib0.readVarUint` reads up to 53 bits via JS `Number`).

### Impact

- **Cross-language interop with `yjs`**: a `yjs` peer that legitimately uses `random.uint53()` for its client ID (the JS default) cannot round-trip its ops through a `yrs` peer — the writer attribution silently changes, breaking convergence.
- **Rust-only deployments at risk**: any `yrs::Doc::with_options { client_id: > u32::MAX }` (the `ClientID = u64` type alias permits this) will silently corrupt every applied update from another such peer.
- **CRDT convergence**: `read_id` truncation re-targets block origins, not just attribution metadata — direct violation of Yjs's "convergent under arbitrary op delivery order" invariant.

### Proposed fix

Two-line change in both sites — read as `u64` directly. PR linked below.

### Why now

We hit this implementing a Rust ⇄ JS interop test in a project we are working on. Our ClientIDs are derived from a per-device key via `BLAKE3_KDF → u64 & ((1<<53) - 1)` — masked to the JS-safe-integer range, so values ≥ 2^32 are overwhelmingly likely. We worked around it locally via a `[patch.crates-io]` vendored fork, but the fix belongs upstream.

### Workaround (no patch)

Constrain client IDs to `[0, 2^32 - 1]` on the encoder side. This loses ~20 bits of collision-resistance budget vs. the `lib0`-compatible 53-bit range and does not interop cleanly with JS yjs peers that use `random.uint53()`.

ravikiran-r:
Closes #609.

### What

Two-line change in `src/updates/decoder.rs`:

- `DecoderV1::read_client` (per-client header, line 143): read as `u64`.
- `DecoderV1::read_id` (block origin / `left_id` / `right_id`, lines 85-88): read `client` as `u64`. `clock` stays `u32` (yrs's clock space is `u32` by design; `StateVector` values are `u32`; `ID::new(client: ClientID, clock: u32)` already takes `u32` for clock).

```diff
@@ DecoderV1::read_id @@
     fn read_id(&mut self) -> Result<ID, Error> {
-        let client: u32 = self.read_var()?;
-        let clock = self.read_var()?;
-        Ok(ID::new(client as ClientID, clock))
+        let client: u64 = self.read_var()?;
+        let clock: u32 = self.read_var()?;
+        Ok(ID::new(client, clock))
     }

@@ Decoder for DecoderV1 :: read_client @@
     #[inline]
     fn read_client(&mut self) -> Result<ClientID, Error> {
-        let client: u32 = self.cursor.read_var()?;
-        Ok(client as ClientID)
+        self.cursor.read_var::<u64>()
     }
```

`DecoderV2` is unchanged — it already reads as `u64`.

### Why

Issue #609 describes the bug: client IDs in `[u32::MAX + 1, u64::MAX]` round-trip losslessly through `lib0.readVarUint` (JS) and through the `yrs` encoder, but `DecoderV1` silently truncates them to the lower 32 bits XOR-rotated (because `read_var<u32>` uses `u32::wrapping_shl` per byte). This breaks Rust ⇄ JS interop and breaks Rust-only deployments that use the full `ClientID = u64` range.

### Behavioural change

For client IDs in `[0, u32::MAX]` the patch is a strict no-op (the previous `read_var<u32>` cast was correct in this range, and `read_var<u64>` returns the same bits). For client IDs in `(u32::MAX, u64::MAX]` the patch unbreaks the round-trip — the value the encoder wrote is now the value the decoder reads.

### Tests

- Adds `tests/clientid_above_u32_round_trips_v1.rs` with two regression gates (one per fixed site): per-client-header round-trip, and a multi-writer merge update where the second writer's op has a left-origin pointing at the first writer's `> 2^32` client ID. Both fail under stock `yrs` and pass under this patch.
- Existing test suite is unaffected (no test in the existing suite happened to exercise client IDs > 2^32, which is why the bug went undetected).

### Compatibility

- Wire format: unchanged. `write_var<u64>` and `read_var<u64>` produce / consume the same byte stream that `lib0` already produces / consumes.
- `Update::blocks.clients` HashMap key type: unchanged (`ClientID = u64`).
- `StateVector::iter()` value semantics: unchanged.
- `yjs` JS peer compatibility: improved (now matches `lib0.readVarUint`'s 53-bit range).

### Discovered via

Working on a Rust ⇄ JS interop test for a project we are working on. We use a per-device BLAKE3-derived `u64 & ((1<<53) - 1)` client ID; values ≥ 2^32 are overwhelmingly common. Our local workaround was a `[patch.crates-io]` vendored fork carrying this exact diff; we'd rather upstream the fix.

### DIFF
--- a/yrs/src/tests/decoder_v1_clientid_above_u32.rs
+++ b/yrs/src/tests/decoder_v1_clientid_above_u32.rs
@@ -1,0 +1,138 @@
+//! Regression gates for issue #609 — `DecoderV1::read_client` and
+//! `read_id` silently truncated clientIDs > 2^32 to u32 via
+//! `read_var_u32`'s `wrapping_shl`. The encoder side and the wire
+//! format both supported the full `ClientID = u64` range; only the V1
+//! decoder dropped the upper bits. `DecoderV2` was always correct.
+//!
+//! These tests fail on the pre-fix DecoderV1 (clientID `0x19266f5beab768`
+//! decodes as `0x5bfbb768` — lower 32 bits XOR-rotated) and pass on
+//! the fixed DecoderV1.
+
+use crate::updates::decoder::Decode;
+use crate::{Doc, GetString, Options, ReadTxn, StateVector, Text, Transact, Update};
+
+/// `DecoderV1::read_client` site — the per-client header in the v1
+/// update binary.
+#[test]
+fn decoderv1_read_client_round_trips_clientids_above_u32() {
+    // 0x19266f5beab768 ≈ 7×10^15 — fits in 53 bits (the JS-yjs safe range)
+    // but well above 2^32 so the bug truncates it.
+    let cid: u64 = 7_079_134_143_100_776;
+    assert!(cid > u64::from(u32::MAX));
+
+    let doc = Doc::with_options(Options {
+        client_id: cid,
+        skip_gc: false,
+        ..Default::default()
+    });
+    let text = doc.get_or_insert_text("content");
+    {
+        let mut txn = doc.transact_mut();
+        text.insert(&mut txn, 0, "x");
+    }
+    let bytes = doc
+        .transact()
+        .encode_state_as_update_v1(&StateVector::default());
+
+    let update = Update::decode_v1(&bytes).expect("decode_v1 must succeed on self-encoded update");
+    let decoded: Vec<u64> = update.state_vector().iter().map(|(c, _)| *c).collect();
+    assert_eq!(
+        decoded,
+        vec![cid],
+        "DecoderV1::read_client must round-trip clientIDs > 2^32 losslessly. \
+         Encoded {cid:#x}, decoded {decoded:?}.",
+    );
+}
+
+/// `DecoderV1::read_id` site — block-origin (`left_id` / `right_id`)
+/// references. More dangerous than `read_client`: corrupts CRDT
+/// ordering, not just attribution metadata. Construct a multi-writer
+/// merge update where the second writer's op has a left-origin
+/// pointing at the first writer's > 2^32 clientID; assert the post-
+/// apply doc's state_vector contains BOTH original clientIDs un-
+/// truncated.
+#[test]
+fn decoderv1_read_id_round_trips_block_origin_clientids_above_u32() {
+    let cid_a: u64 = 7_079_134_143_100_776; // 0x19266f5beab768
+    let cid_b: u64 = 6_332_093_782_882_047; // 0x167f01789cf6ff (distinct, > u32::MAX)
+    assert!(cid_a > u64::from(u32::MAX) && cid_b > u64::from(u32::MAX));
+    assert_ne!(cid_a, cid_b);
+
+    // Step 1: doc_a inserts "a".
+    let doc_a = Doc::with_options(Options {
+        client_id: cid_a,
+        skip_gc: false,
+        ..Default::default()
+    });
+    let text_a = doc_a.get_or_insert_text("content");
+    {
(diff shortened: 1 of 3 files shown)


Horusiath:
Closing this one, as the issue was addressed by #612 .

- This PR looks like it was generated by AI without any proper care. 
- It doesn't address compatibility issues (you'd know if you have run the tests) with existing users which was the major reason why we postponed ClientID change. #612 provides an escape hatch to keep the previous behaviour.

PR #617, opened 2026-05-14
Remove move feature

Horusiath:
As part to process to move with parity with yjs v14, we're going to remove the `move_to`/`move_range` methods. This is dictated by the high maintenance and complexity cost, while bringing a little on the table for a small amount of users.


SylvainCorlay:
This is really unfortunate as we funded work to support move operations in yjs, but are still unable to use them because it was never included in a release. Landing this eventually would be key for Jupyter UX.

Horusiath:
@SylvainCorlay I've described on how to implement such feature in a simpler form that shouldn't require heavy interference into yrs/yjs internals: https://www.bartoszsypytkowski.com/replacing-yjs-move-feature/

dmonad:
Hey @SylvainCorlay,

I'm sorry that this feature was removed. The work on the range-move feature in Yjs (which is distinct from y-crdt) was funded by QuantStack. The feature was supported for a time in Yjs and was eventually removed to reduce maintenance burden.

I understand that this is unfortunate for the Jupyter Project. I remember that this was used to retain the undo history when cells are moved. The range-move feature is quite ambitious and touches nearly every aspect of Yjs/Y-CRDT. Removing it was the right decision for us to support attribution of content, as it reduces maintenance burden significantly.

Bartosz implemented a less ambitious move feature (single-item move) that I'd like to support in Yjs as well eventually. I highly recommend reading Bartosz's article about a non-native move feature. But you could also look into using CodeMirror's native undo manager.

The critique should be directed at me. While I'm under no legal obligation to maintain funded work, I feel bad for leaving JupyterLab no alternative but to remove the cell UndoManager.

I understand your frustration with the removal of this feature in Yjs, but it's also worth noting that y-crdt was never funded by QuantStack (nor was the implementation of the move feature in y-crdt). The y-collective actually funded some the early Y-CRDT and Ypy work. Bartosz provides free maintenance for Y-CRDT and supports the development of pycrdt for free as well. He is under absolutely no obligation to implement, support, or maintain the move feature - morally or otherwise.

SylvainCorlay:
Hi @dmonad

Of course, nobody is bound to maintain funded work forever! Though I presume it is an interesting data point to maintainers that we still care about the feature in downstream projects! On a side note to the point, I think we are engaging really positively. Our investment in the YJS stack has continued beyond the funding of core developers from a few years ago. QuantStack team members still maintain pycrdt, and another one is actively working on R bindings.

Just like you, a considerable part of our maintenance work in Jupyter and other projects is not funded by clients - and we sometimes have to make difficult calls on maintaining barely used features - so I totally get why such calls would be made.

We will definitely read about the proposed alternative.

dmonad:
> On a side note to the point, I think we are engaging really positively.

I feel the same way!

PR #622, opened 2026-06-03
release v0.27

Horusiath:
- New type: `IdMap` (#616) - works similar to `IdSet` but allows for attribution of individual ranges it describes.
- `UndoManager` now works across multiple documents (#620, #621):
    - **breaking**: `UndoManager::new` and other constructors don't take initial doc/type anymore.
    - **breaking**: `UndoManager::expand_scope` now required `&Doc` parameter (since it can operate over multiple documents).
    - **breaking**: `StackItem.doc` field to define `Doc::guid` of the document the stack item refers to.
    - **breaking**: `UndoManager::clear` method has been replaced by more granular `clear_undo`, `clear_redo`, `clear_all` methods.
    - New API: `UndoManager.stack_cleared` event handler, 
- Removed `move_to`/`move_range` related API (#617). Article describing workaround: https://www.bartoszsypytkowski.com/replacing-yjs-move-feature/
- Observer callbacks now accepts `FnMut` instead of `Fn` (#619)
- Transaction related logic has been updated to match yjs v14 (#618):
    - introduction of `Transaction.insert_set`.
    - `Transaction.local`
    - support for cleanup formatting
    - now it's possible to integrate out-of-order updates if they don't have dependency over missing data

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1257,7 +1257,7 @@ dependencies = [
 
 [[package]]
 name = "yffi"
-version = "0.26.0"
+version = "0.27.0"
 dependencies = [
  "serde_json",
  "yrs",
@@ -1265,7 +1265,7 @@ dependencies = [
 
 [[package]]
 name = "yrs"
-version = "0.26.0"
+version = "0.27.0"
 dependencies = [
  "arc-swap",
  "assert_matches2",
@@ -1289,7 +1289,7 @@ dependencies = [
 
 [[package]]
 name = "ywasm"
-version = "0.26.0"
+version = "0.27.0"
 dependencies = [
  "base64_light",
  "console_error_panic_hook",
--- a/tests-wasm/package-lock.json
+++ b/tests-wasm/package-lock.json
@@ -16,7 +16,7 @@
     },
     "../ywasm/pkg": {
       "name": "ywasm",
-      "version": "0.26.0",
+      "version": "0.27.0",
       "license": "MIT"
     },
     "node_modules/isomorphic.js": {
--- a/yffi/Cargo.toml
+++ b/yffi/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "yffi"
-version = "0.26.0"
+version = "0.27.0"
 authors = ["Kevin Jahns <kevin.jahns@protonmail.com>", "Bartosz Sypytkowski <b.sypytkowski@gmail.com>"]
 keywords = ["crdt", "c-ffi", "yrs"]
 edition = "2018"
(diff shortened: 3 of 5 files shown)


davidbrochart:
There are still mentions of `try_undo` and `try_redo` in the undo manager documentation, which have been removed.

davidbrochart:
It would have been nice to keep `try_undo` and `try_redo` if we don't want to deadlock trying to acquire a transaction.

Horusiath:
@davidbrochart Technically you can use regular `UndoManager::undo`/`redo` with similar effect, as the only awaitable element of that method is about waiting for transaction lock to be released. You can just [poll](https://docs.rs/futures/latest/futures/macro.poll.html) (or [poll_once](https://docs.rs/futures-lite/latest/futures_lite/future/fn.poll_once.html)) returned future once and check for the result:

```rust
use std::task::Poll;
use futures::poll;

let mut um = UndoManager::new();
let future = um.undo(); 
match poll!(future) {
    Poll::Ready(_) => // successfully acquired transaction and finished the operation,
    Poll::Pending => // awaiting for transaction lock to be released,
}
```

davidbrochart:
Alternatively, can we use `doc.try_transact_mut()`?

Horusiath:
Unfortunately, undo/redo acquires transaction on demand for serving `UndoManager` only: now it's even more tricky, since undo StackItem keeps the data about which document it's related to this specific undo item, so you don't really know which Doc's transaction you may need in advance (given you use multiple docs on single undo manager).

I can add `try_undo`/`try_redo` if my proposed workaround is a blocker for you.

davidbrochart:
> I can add `try_undo`/`try_redo` if my proposed workaround is a blocker for you.

IIUC `poll!(future)` can only be called inside an async function? If so, this is be a blocker for me.

Horusiath:
No, other way around: `poll!` is used to call asynchronous function in synchronous context. Also, I personally recommend `futures-lite` crate which is more lightweight and `poll_once!` is more straightforward to achieve what you're trying to check.

davidbrochart:
It doesn't seem to be trivial, the only examples I found imply using a waker and a context, and using `pin!`:
```rust
    pub fn redo(&mut self)  -> PyResult<bool> {
        let mut future = pin!(self.undo_manager.redo());
        let waker = noop_waker();
        let mut cx = Context::from_waker(&waker);
        match future.poll(&mut cx) {
            Poll::Ready(value) => { Ok(value) }
            Poll::Pending => { Err(PyRuntimeError::new_err("Cannot acquire transaction")) }
        }
    }
```
And `poll!` is only usable inside of async functions (see [here](https://docs.rs/futures/latest/futures/macro.poll.html)).
Do you have a better way?

Horusiath:
If that's an issue, I'll include regular try_undo in the next release.

davidbrochart:
I don't know if that's an issue, is the code above correct or is there a simpler way?
In any case, I imagine I won't be the only one who needs `try_undo`/`try_redo` so it's worth adding them back.

Horusiath:
Your snippet should be fine.

As for `try_` methods. I'm looking for the way to get rid of locks from yrs entirely. Possibly move them to caller space (where user who wants to use Doc in multi-threaded context is responsible for lock maintenance). There are many lock libraries with different tradeoffs and all of them are relatively heavy, at least in binary size (which matter i.e. in the browser).

Issue #415, opened 2024-04-10
use errors instead of panic


zignis:
There are numerous instances of panic in the codebase, which are not ideal for production use cases. This often results in crashes of the processes on our production machines because we accept dynamic documents from users, which might contain malformed data. I believe that returning errors instead of panicking would be a better approach.

For instance, the following line has been causing panics in our production instance:

https://github.com/y-crdt/y-crdt/blob/567f83da810c146f9b878cb8dccea3c25468b6a0/yrs/src/slice.rs#L239

We cannot always verify whether the incoming data from the client is correct.

Horusiath:
@zignis we're at the process of removing panics. However the ones that usually remained points to potentially bugs or configuration issues. The fact that you've run into this error most likely points to one of these issues.

zignis:
@Horusiath This happened randomly with one of our users, and I cannot provide their document due to obvious reasons. I am trying to reproduce it from a new document, and I'll get back to you if I succeed.

appflowy:
@Horusiath  I have the same issues but I don't how to reproduce them. It happened when i try to call
`txn.encode_state_as_update_v1(&StateVector::default())`

weironz:
@Horusiath — here's the reproducible scenario you asked for, minimal and built entirely from the public API on **yrs 0.27.3** (rustc 1.97.1). A single-byte corruption of an otherwise-valid v1 update reaches two distinct failure modes through the safe `Update::decode_v1` + `Transaction::apply_update` path:

```rust
use yrs::updates::decoder::Decode;
use yrs::{Doc, ReadTxn, StateVector, Text, Transact, Update};

fn main() {
    // A valid v1 update produced by yrs itself (52 bytes).
    let good = {
        let doc = Doc::new();
        let text = doc.get_or_insert_text("t");
        text.push(&mut doc.transact_mut(), "hello world, this is a yrs document");
        let txn = doc.transact();
        txn.encode_state_as_update_v1(&StateVector::default())
    };

    // (1) Flip byte 9 -> panic at yrs-0.27.3/src/block.rs:92
    //     "assertion failed: value & Self::MASK == 0"
    let mut bad = good.clone();
    bad[9] ^= 0xff;
    if let Ok(u) = Update::decode_v1(&bad) {
        let _ = Doc::new().transact_mut().apply_update(u);
    }

    // (2) Flip byte 50 (string content) -> invalid UTF-8 read:
[...]

PR #644, opened 2026-08-04
fix: validate UTF-8 when decoding string content from the wire

weironz:
Rebased onto `main` and narrowed: the allocation half of this PR is now redundant with #639 (merged), so it has been dropped. What remains is a different failure class that #639 does not touch.

Follow-up to my comment on #415 with the reproducers.

Context: we run a server that accepts yrs updates over the wire from authenticated clients, so malformed input is reachable in practice — a client bug or a corrupted on-disk cache is enough, no malice required. We wrap the calls in `catch_unwind`, which handles the panic class but structurally cannot catch a non-unwinding abort.

### Unchecked UTF-8 — `encoding/read.rs` `read_string`, `updates/decoder.rs` `StringDecoder::new`

```rust
Ok(unsafe { std::str::from_utf8_unchecked(buf) })
```

String content arrives from the wire, so building a `&str` from it without validating is UB on malformed input — a non-unwinding abort in debug (`invalid value for char` / `unreachable_unchecked must never be reached`), silently unsound in release. Fixed with checked `from_utf8`. Cost is one validation pass over string content per decode.

This is orthogonal to #639. That PR was about *fallible allocation*: turning an attacker-controlled length prefix into a recoverable `Error::NotEnoughMemory` instead of `handle_alloc_error`. Neither `try_reserve` nor a caller's `catch_unwind` does anything about constructing a `&str` from bytes that are not UTF-8.

The third `from_utf8_unchecked` (`lib.rs:777`) is over a HEX table that is always valid ASCII — left alone.

### Tests

Added to `update.rs`: the UTF-8 reproducer, plus a sweep asserting that no single-byte corruption of a valid update aborts. The sweep deliberately tolerates unwinding panics and fails only on aborts — a caller can contain a panic, and the remaining ones are tracked in #415; it is the abort class that has to be closed in the library. (The sweep currently trips the debug-only `ClientID::new` assert at `block.rs:92`, compiled out in release, which is why it is wrapped rather than asserted against.)

Reverting the fix makes the new test abort.

---

*Prepared with AI assistance; reviewed, tested, and verified by the submitter, who takes responsibility for this contribution.*


Horusiath:
#639 seems to cover the same issue

montfort:
On the duplicate call: #639 (now merged) covers the first half of this PR, but not the second.

**Covered.** `Any::decode` cases 117/118 in `any.rs` — #639 landed the same `try_reserve` fix there, so that part is genuinely a duplicate.

**Not covered.** The two `from_utf8_unchecked` sites are untouched by #639 and are still on `main` today:

- `yrs/src/encoding/read.rs:137` (`read_string`)
- `yrs/src/updates/decoder.rs:486` (`StringDecoder::new`)

These are a different failure class, not the same issue. #639 was strictly about *fallible allocation*: turning an attacker-controlled length prefix into a recoverable `Error::NotEnoughMemory` instead of `handle_alloc_error`. Building a `&str` from unvalidated wire bytes is UB — an abort in debug, silently unsound in release — and neither `try_reserve` nor a caller's `catch_unwind` does anything about it.

So closing #644 as a duplicate would take the UTF-8 half with it.

For context on why I care beyond the overlap: we run `yrs` behind a C-ABI shim where every entry point is wrapped in `catch_unwind`, precisely because a panic crossing the FFI boundary is UB. That guard is structurally blind to the abort class — which is what motivated #639, and as far as I can tell applies just as much to the UTF-8 sites here.

weironz:
You're right that #639 overlaps — but only on half of this. I've rebased and dropped that half, so the diff now shows the difference rather than arguing about it.

**Dropped:** `Any::decode` cases 117/118 in `any.rs`. #639 landed the same `try_reserve` fix, and went further by covering `id_set.rs`, `state_vector.rs` and `awareness.rs` as well. `git diff` against `main` for that file is now empty, and the reproducer that went with it is gone too.

**What's left** is a different failure class that #639 doesn't touch — two `from_utf8_unchecked` sites still on `main` today:

- `yrs/src/encoding/read.rs:137` (`read_string`)
- `yrs/src/updates/decoder.rs:486` (`StringDecoder::new`)

#639 was about *fallible allocation*: turning an attacker-controlled length prefix into a recoverable `Error::NotEnoughMemory` instead of `handle_alloc_error`. Building a `&str` from unvalidated wire bytes isn't an allocation problem at all — it's UB, which aborts in debug and is silently unsound in release. Neither `try_reserve` nor a caller's `catch_unwind` addresses it.

The PR is now 3 files, +76/−2: the two checked `from_utf8` conversions, the UTF-8 reproducer, and a sweep asserting no single-byte corruption of a valid update aborts. It also merges cleanly again. Reverting the fix makes the new test abort.

adstep:
Independently discovered this unsafe decoding too. Also worried about this from the perspective of having the updates come over the wire from users so the input being inherently untrusted.

Solution aligns with what I was expecting.

This seems like it could have broad impact on a lot of people publicly hosting this?

Horusiath (changes requested):
There are failing tests in your PR. Also: please don't include comments that you have not written. This wall of text in tests could be one sentence.

Horusiath (approved):

PR #645, opened 2026-08-05
fix: guard against divide-by-zero in find_index when end == 0

petergaultney:
A single-block client stream where that block's own range is (0, 0) - a length-1 block at clock 0 - causes a panic at `clock / end` on line 58 of `block_store.rs`. This happens in production with few-edit documents, which produce exactly this single-block stream shape.

Since `clock != start` was already checked on line 54, the requested clock is out of range and returning `None` is correct.


Horusiath:
Thanks for your PR, but the fix is not complete and doesn't include reproducible tests. I've included fixes in #647

petergaultney:
Your fix still doesn't guard against underflow. 

try `[(10,14),(15,19)] with clock=2` and you'll see what I mean. ❤️

Horusiath:
@petergaultney can you actually create a reproducible scenario? Even with out-of-order application of updates, you shouldn't be able to create a case, when block list is not empty and has a hole covering index 0 in it. I tried to recreate such case, but it doesn't seem to be reachable situation.

PR #648, opened 2026-08-17
fix: find_index underflows when clock precedes the first block

petergaultney:
Follow-up to #647, which fixed the case where `clock` lands past the last block. The mirror
case still panics.

When `clock` precedes the first block's start, the search walks left. At `mid == 0`,
`right = mid - 1` wraps on `usize`, and the next iteration indexes far out of range:

```rust
} else {
    right = mid - 1;   // mid == 0 here wraps to usize::MAX
}
```

`find_index(&[(10,14), (15,19)], 2)` panics with `attempt to subtract with overflow` in debug,
and indexes out of bounds in release.

## Where it's reachable

Anywhere a queried clock can precede the local block list. `apply_delete` takes its range
straight from a remote peer's delete set (`transaction.rs:648`), so a peer deleting content
whose early blocks the receiver never received, or has GC'd, reaches it.

Two things suggest this shape was already known and handled per-caller rather than centrally:
`store.rs:234` clamps with `.max(first block's clock_start)` and comments "make sure the first
id exists", and `store.rs:182` queries `clock - 1`.

Found in production, on a per-span authorship walk over freshly-created documents. The panic
was caught upstack and the content silently returned unattributed, which is how it went
unnoticed for a while.

## The change

A `mid == 0` guard returning `None`, plus a test module for `find_index`. The test covers the
new case and the two `#647` already handles, so the whole boundary stays pinned.

Without the guard, `clock_below_the_first_block` panics; the other two pass. Full `yrs` lib
suite: 383 passed, 0 failed, 1 ignored.

## Note on verification

`upstream/main` does not build on current stable Rust (1.94.1) - `block.rs:1019` uses an
`if let` guard, still unstable. I ran the suite with `RUSTC_BOOTSTRAP=1
-Zcrate-attr=feature(if_let_guard)` to get past it. Unrelated to this change, but you may want
to know if stable builds are meant to work.

This supersedes #645, which you closed in favour of #647. #647 covers two of the three cases
#645 raised; this is only the remaining one.


yrs/src/block_store.rs
```diff
@@ -502,3 +506,46 @@ impl std::fmt::Display for BlockStore {
...
+                        ID::new(ClientID::new(1), *start),
+                        end - start + 1,
+                    )))
+                })
+                .collect(),
+        }
+    }
+
+    #[test]
+    fn clock_below_the_first_block() {
```
Horusiath, line 533: Please don't use internals to point out the issue. Try to replicate the actual scenario of Doc and operations/updates, where such structure is being constructed instead. Only this way we can be sure that the fix we're applying aims to fix the state that can be produced in reality.

Horusiath (changes requested):

Horusiath:
As for `if let` guard - [it became stable in Rust 1.95](https://releases.rs/docs/1.95.0/), released in April this year.

PR #652, opened 2026-08-27
fix duplicate typedefs in generated C header

adstep:
## What

`yffi/cbindgen.toml` hand-declares its opaque handle types in the `header` preamble and uses `[export] exclude` to stop cbindgen re-emitting them. Three exclusions never took effect, so the generated header ships self-referential typedefs:

```c
typedef YDoc YDoc;
typedef Branch Branch;
typedef YSubscription YSubscription;
```

## How this surfaced

I hit this while evaluating the C API from Go. cgo parses the header rather than compiling it, so it rejects outright what a C compiler waves through:

```go
package main

/*
#cgo CFLAGS: -I/path/to/y-crdt/tests-ffi/include
#include <libyrs.h>
*/
import "C"

func main() { var _ *C.YDoc }
```

```sh
go mod init cgo-check && go vet .
# on main: cgo: ./main.go:9:22: type conversion loop at YDoc
```

## Why it happens

cbindgen runs `remove_excluded()` before `rename_items()`, and the filter matches the original Rust identifier, so `exclude` entries must use pre-rename names.

| Rust type | rename | in `exclude`? | before |
|---|---|---|---|
| `Doc` | -> `YDoc` | absent | duplicate |
| `Branch` | none | absent | duplicate |
| `Subscription` | -> `YSubscription` | listed post-rename | never matched -> duplicate |

## Verification

The defect on `main`, with no Go involved. 3 errors, exit 1:

```sh
git show main:tests-ffi/include/libyrs.h | clang -x c -std=c99 -pedantic-errors -Wno-gnu-empty-struct -fsyntax-only -
```

```
error: redefinition of typedef 'YDoc' is a C11 feature [-Werror,-Wtypedef-redefinition]
error: redefinition of typedef 'Branch' is a C11 feature [-Werror,-Wtypedef-redefinition]
error: redefinition of typedef 'YSubscription' is a C11 feature [-Werror,-Wtypedef-redefinition]
```

The same command against this branch exits 0. (`-Wno-gnu-empty-struct` suppresses 15 pre-existing warnings unrelated to this change; under `-std=c11` both headers pass, which is why CI never caught it.)

The `tests-ffi` suite passes unchanged: 39/39 cases, 368/368 assertions. No Rust source is touched and `nm` over the built `libyrs.dylib` is identical before and after, so the ABI is unaffected.


Horusiath (approved):
Thanks for noticing! 👍

PR #655, opened 2026-08-29
fix: release observer callbacks when Subscription is dropped

### LINKED ISSUE #654: Dropping Subscription retains its callback until future observer activity, causing leaks
## Summary

Since `yrs` 0.27.0, dropping a `Subscription` does not immediately release its callback.

`Subscription::drop` queues the callback's `Origin` in `PendingRemovals`, but the callback remains in `Observer::callbacks` until a later `subscribe`, `subscribe_with`, or `trigger` call drains the queue.

If the observer becomes idle, that cleanup may never happen.

As a result:

* Resources captured by the callback remain retained after the subscription is dropped.
* `has_subscribers()` continues to report the dropped subscription as active.
* If the callback captures an object that transitively owns the observer, the retained callback can create an ownership cycle and keep the object alive indefinitely.

This also appears inconsistent with the documented behavior of `Subscription`, where dropping it unsubscribes the corresponding callback.

## Reproduction

A binding-free regression test can reproduce the issue:

```rust
use std::sync::Arc;
use yrs::observer::Observer;

#[test]
fn dropping_subscription_releases_callback() {
    let mut observer: Observer<Box<dyn FnMut() + Send + Sync + 'static>> =
        Observer::new();

    let captured = Arc::new(());
    let weak = Arc::downgrade(&captured);

    let subscription = observer.subscribe(Box::new(move || {
        let _ = &captured;
    }));

    assert!(weak.upgrade().is_some());
    assert!(observer.has_subscribers());

    drop(subscription);

    assert!(weak.upgrade().is_none());
    assert!(!observer.has_subscribers());
}
```

Both final assertions currently fail on `main`.

This should be tested with both the default and `sync` feature configurations.

## Regression

The behavior appears to have been introduced by https://github.com/y-crdt/y-crdt/commit/764ebea39e56256ce4108ed719ce75875fcf0426 in PR #619, which added support for `FnMut` observer callbacks.

Before that change, dropping a subscription removed its callback immediately. After the change, removal became deferred through `PendingRemovals`.

## Expected behavior

Dropping a `Subscription` should make the subscription inactive and release its callback without requiring future observer activity.

If a callback is already executing when its subscription is dropped, its captured state should become releasable after that invocation finishes.

This should remain safe when a subscription is dropped:

* from inside its own callback;
* from another callback during the same dispatch;
* with `sync`, from another thread while dispatch is in progress.

`has_subscribers()` should also stop counting dropped subscriptions, while preserving `FnMut` support.

## Scope

The affected path is `Observer::subscribe`, where the returned `Subscription` is responsible for unregistering the callback when dropped.

`Observer::unsubscribe(&Origin)` removes keyed callbacks directly, so this specific deferred-drop behavior does not apply to explicit keyed unsubscription.

## Affected versions

Confirmed in:

* `v0.27.0`
* `v0.27.1`
* `v0.27.2`
* `v0.27.3`
* `v0.27.4`
* current `main`

The regression is not present before https://github.com/y-crdt/y-crdt/commit/764ebea39e56256ce4108ed719ce75875fcf0426.

## Downstream impact

We encountered this through [pycrdt](https://github.com/jupyter-server/pycrdt), where a callback can capture the Python `Doc` while the document transitively owns the Rust observer. If the callback remains retained after `unobserve()`, that cross-language ownership cycle keeps the document alive.

The downstream investigation is documented in [jupyter-server-documents#270](https://github.com/jupyter-ai-contrib/jupyter-server-documents/issues/270).

ebrattli:
## Summary

Since 0.27.0, dropping a `Subscription` does not release its callback. Removal is queued in `PendingRemovals` and only applied on a later `subscribe`, `subscribe_with`, or `trigger`. If the observer goes idle, that cleanup never runs.

Captured state then stays alive, `has_subscribers()` still reports the handle as active, and a callback that captures a `Doc` (or anything that owns the observer) can pin that object forever. We hit this in pycrdt: `unobserve()` left a Rust callback holding the Python `Doc`.

`PendingRemovals` cannot be patched in place while keeping `FnMut`. `trigger` borrows the callback, so drop cannot remove it from the map and must not block on an in-flight call (a callback can drop its own subscription).

This change stores each callback in a `Ready | Running | Cancelled` slot. `Subscription` holds a weak handle, so drop cancels immediately, takes `F` out of the slot lock before user destructors, and skips later callbacks in the same `trigger` that were already cancelled. `FnMut` is unchanged.

## Related issues

Fixes #654
Regression from #619
Downstream: jupyter-server/pycrdt and jupyter-ai-contrib/jupyter-server-documents#270

## How to test

```bash
cargo test -p yrs --lib observer
cargo test -p yrs --lib observer --features sync
```

The new tests cover immediate release, `has_subscribers()`, self-unsubscribe during the callback, skipping a sibling cancelled in the same dispatch, panic restore, replace/unsubscribe dropping the old callback, and observer drop breaking handle cycles.

CI already runs default and `--all-features` on yrs.

## Notes

- `has_subscribers()` becomes false as soon as the last subscription drops, without waiting for a later `trigger`.
- A callback cancelled by an earlier one in the same `trigger` is skipped. Previously it still ran because removal was only queued.
- The callback destructor runs on the thread that drops `Subscription`, unless it was in-flight, in which case it runs on the dispatch thread after that call returns.
- `Observer::subscribe` now requires `F: Send + 'static` with `sync`, and `F: 'static` without. Higher-level `observe()` already had these bounds; direct `Observer` users may need to add them.
- Empty slot shells can remain after an anonymous subscribe/drop until the next `subscribe`, `subscribe_with`, `unsubscribe`, or `trigger`. The callback payload is already gone. `unsubscribe` and keyed replace compact immediately.


Horusiath:
You know that you can provide a key to subscribe/unsubscribe manually? That's how `ywasm` JS plugin works for example.

ebrattli:
I had a look at observe_with. That would solve the leak, but it has another issue. If you unsubscribe from inside a callback, that can be undefined behavior: `Observer::trigger` iterates `callbacks.values_mut()`, and unsubscribe mutates that same HashMap. It can even drop the `Box<dyn FnMut>` that is currently running. It looks like `Subscription::drop` was deferred to avoid this, but keyed unsubscribe never got the same treatment.

Unsubscribing inside a callback is a pattern used in pycrdt for the Async iterator.

PR #659, opened 2026-09-16
docs(ywasm): update stale docs about UTF-8 offsets

kavinsood:
The ywasm documentation previously claimed `YText` and `YXmlText` offsets were UTF-8 bytes. 

This is completely false. The underlying JS API expects UTF-16 code units. Relying on the old documentation causes silent document corruption the second a user inserts a multi-byte Unicode character (emojis, CJK, etc.).

Fix the docstrings to reflect reality and prevent people from destroying their CRDT graphs.

## Summary

- correct the `YText` and `YXmlText` API documentation to describe JavaScript-visible positions and lengths as UTF-16 code units
- align the binding documentation with JavaScript string indexing and Yjs behavior
- cover the type-level, `length`, and `delete` documentation in both bindings

## Validation

- `cargo test -p ywasm --lib`
- `git diff --check`

`cargo fmt --all -- --check` currently reports pre-existing formatting differences in unrelated workspace files; this PR does not change them.


Horusiath (approved):
Thanks for noticing.

PR #660, opened 2026-09-16
feat(ywasm): Expose applyUpdateAndCheckIfChanged to avoid O(N) diff allocations

kavinsood:
Previously, the only way to know if an applied update actually changed the document was to encode the state vector before and after applying, and compare them. This is horribly inefficient and allocates unnecessary buffers. Worse, it fails on delete-only updates because deletions don't advance the state vector.

This patch exposes the `internal has_update_changes()` predicate to the Wasm boundary, returning a simple boolean. This allows hosts to efficiently drop duplicate remote updates and correctly identify delete-only updates without expensive buffer allocations.

## Validation

- `cargo test -p yrs update_change_detection_covers_duplicate_and_delete_only_updates --lib`
- `cargo test -p ywasm --lib`
- `git diff --check`

The regression covers insertion, duplicate insertion, delete-only application (whose state vector is unchanged), and duplicate deletion.

`cargo fmt --all -- --check` currently reports pre-existing formatting differences in unrelated workspace files; none are in the files changed by this PR.


Horusiath:
I don't see how it's supposed to help you. You're talking about change in context of `ywasm`, yet no method was exposed there.

kavinsood:
> I don't see how it's supposed to help you. You're talking about change in context of `ywasm`, yet no method was exposed there.

It is exposed at the bottom of the diff in `ywasm/src/lib.rs`, line 337.

```rust
#[wasm_bindgen(js_name = applyUpdateAndCheckIfChanged)]
pub fn apply_update_and_check_if_changed(
    doc: &Doc,
    update: js_sys::Uint8Array,
    origin: JsValue,
) -> Result<bool> {
// ...
    txn.apply_update(update)
        .map_err(|e| JsValue::from(e.to_string()))?;
    Ok(txn.has_update_changes())
}

```

It directly exposes the predicate to the JS host, returning the boolean in a single Wasm crossing.

PR #664, opened 2026-09-17
Preserve large integer types in binary and serde encoding

davidbrochart:


yrs/src/any.rs
```diff
@@ -145,7 +145,12 @@ impl Any {
                 match *num {
                     Number::Int(n) => {
                         // ensure max compatibility with yjs lib0 number encoding
-                        if n >= -0x7FFFFFFF && n <= 0x7FFFFFFF {
+                        if !(Number::I64_MIN_SAFE_INTEGER..=Number::I64_MAX_SAFE_INTEGER).contains(&n) {
```
Horusiath, line 148: The last else branch of this match case already handles BigInt. The reason why it's this way is to match yjs behaviour.

If going through `Number::Float` variant during decoding is a problem: you shouldn't match against `Number` directly anyway. Similar to how serde_json uses `as_i64`/`as_f64` when decoding, there are similar methods here so you can convert number to an int/float depending on your need, and they already cover both Number cases.

Horusiath (commented):
What's the purpose of this PR?

davidbrochart:
> What's the purpose of this PR?

It seems that in Yrs v0.28 an integer such as `2**53` is encoded as a float, so we loose the type information after decoding. It's a breaking change and all I could find in the release notes is:
- `Any::Int`/`Any::Float` have been replaced by single `Any::Number` with `as_i64`/`as_f64` converters

But it doesn't mention this new behavior. Your [comment](https://github.com/y-crdt/y-crdt/pull/664#discussion_r4043355235) seems to indicate that it's to match Yjs's behavior, but it breaks pycrdt's existing Python types: previously, large Python integers survived synchronization as integers, while floats remained floats. Using `as_i64()` recovers the numeric value, but cannot distinguish an originally supplied integer from an integral float once both have the same encoding.
I understand that preserving this distinction may conflict with the intended Yjs-compatible behavior though. Separately, there seems to be a precision bug: the float-to-integer cast in the encoding check saturates, so `i64::MAX` (`2**63-1`) passes the check despite rounding to `2**63` as a float.

Horusiath:
You can always try to convert to `i64` first and in case of failure try `f64`. As for the precision bug: I'll make a regression test and make a bug fix.

davidbrochart:
> You can always try to convert to `i64` first and in case of failure try `f64`.

But we still loose the original type.

Horusiath:
Do you interact with Yjs as well? How do you know the original type then?

AFAIK, you when working with other languages you have one of two options:
1. You have dynamic type and can accept anything. That's why JavaScript has no difference between integer and float numbers.
2. Type is part of the contract, so by deserialising the type's field you already know what to expect. That's how serialisation of `Any` type in serde works: if the field is of integer type, we just use `as_i64` and it will coerce to integer (if possible) regardless if internal encoding was `Number::Int` or `Number::Float`.

davidbrochart:
> Do you interact with Yjs as well? How do you know the original type then?

No, just `pycrdt`: a value of `2**53` is a Python `int` in the original document, but becomes a `float` after syncing to another document. Before Yrs v0.28, that value uses the `BigInt` path and comes back as an `int`.

